PR #16, opened 2011-10-17
add auth hooks

ccyphers:
Adding --auth_hook option.  This allows
one to filter request on the initial handshake
based on the JS sending an additional string in in the
protocol list:

websocket = new WebSocket(uri, ['base64', user + '-----' + auth])

This allows one to host the html/js under one type of appserver, integrating into the rest of their site's content and to set the two cookies fed to WebSocket.  Then whtn the ws or wss handshake takes place, websocket.py, can call the auth_hook and disconnect if not authorized.  The --auth_hook is simply an external application the user writes which exits with 0 if allowed.

### DIFF
--- a/websocket.py
+++ b/websocket.py
@@ -16,11 +16,11 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import os, sys, time, errno, signal, socket, traceback, select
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Closing due to no response.

For future reference: this can be implemented in terms of #162

PR #54, opened 2012-08-28
Allow sourcing data from stdin/stdout

engie:
These changes allow sourcing data from stdin to share out to a websocket client. For example:
    ./gen_data.sh | ./websockify 1234
To allow this:
* Logging messages is sent to stderr
* If websockify is using stdin/stdout (target_local = true) then only allow one client to connect at a time


DirectXMan12:
closing due to no response

PR #58, opened 2012-09-12
Add authentication/firewall capability through shell call/args

mightypenguin:
This is based on Cliff Cypher's work. (https://github.com/ccyphers/websockify/commit/852de017686dd930399252ac8f1cda00a15a7a4e)
Let me know what you think. I believe I addressed your notes on his commit request.

I included an example script 'auth.py' to show how token based authentication would work.
This seems to be working fairly well atm.

example call:
./websockify --auth-hook='python auth.py' 10000 host:5900

### DIFF
--- a/auth.py
+++ b/auth.py
@@ -2,14 +2,14 @@
 
 import os
 
+#for k,v in os.environ.items():
+#    print k + " = " + v
+
 try:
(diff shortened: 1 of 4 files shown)


DirectXMan12:
Closing.  You can do something similar with token hooks (#162), and this PR is quite old.

Issue #87, opened 2013-07-24
Enhancement - Replace print() logging with Python logging


rohitkode:
Currently all logging is done via the print function, which writes to console by default.
If a third party application that has it's own logging subsystem uses the websockify library, some logs would be sent to the application's logger (syslog/file etc.) and others to console. 
It would be helpful if websockify used python's logging system, so that log messages can be formatted, logged in different Files,Syslog etc.
The third party application should also be able to control websockify's logging levels.

PR #88, opened 2013-07-24
Add logger for websockify

nttdatainc-openstackers:
Replace print() statement logging with python logging

Addresses issue #87


dwang2:
Can we have websockify logger write to a file through command options?

DirectXMan12:
@dwang2 Please open a separate issue for this

PR #114, opened 2013-12-27
fix cygwin compilation

r1b:
When compiling under cygwin, access() constants are not defined in fcntl.h:

```
websockify.c: In function ‘main’:
websockify.c:368:35: error: ‘R_OK’ undeclared (first use in this function)
         if (access(settings.cert, R_OK) != 0) {
```

Including unistd.h allows compilation to complete.

### DIFF
--- a/other/websockify.c
+++ b/other/websockify.c
@@ -16,6 +16,7 @@
 #include <netdb.h>
 #include <sys/select.h>
 #include <fcntl.h>
+#include <unistd.h>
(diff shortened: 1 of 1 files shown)


r1b:
Bump - 4 years later a conditional include seems more sensible ☺️

CendioOssman:
Probably shouldn't even be conditional. `unistd.h` is the documented location for `access()`. :)

PR #126, opened 2014-04-28
Show main exception only with --verbose

lzap:
We call websockify from our web app and users often hit used port. In that case
websockify exceptions out with strack trace which causes ABRT (Bug Reporting
Tool) to collect exception information, system information and making a report.

This patch prints error messages by default to stderr and only if --verbose is
used, the full stacktrace is shown. I can introduce another option like --trace
if you do not like this one.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -402,21 +402,27 @@ def websockify_init():
         try:    opts.target_port = int(opts.target_port)
         except: parser.error("Error parsing target port")
 
-    # Transform to absolute path as daemon may chdir
-    if opts.target_cfg:
-        opts.target_cfg = os.path.abspath(opts.target_cfg)
+    try:
+        # Transform to absolute path as daemon may chdir
+        if opts.target_cfg:
+            opts.target_cfg = os.path.abspath(opts.target_cfg)
 
-    # Create and start the WebSockets proxy
-    libserver = opts.libserver
-    del opts.libserver
-    if libserver:
-        # Use standard Python SocketServer framework
-        server = LibProxyServer(**opts.__dict__)
-        server.serve_forever()
-    else:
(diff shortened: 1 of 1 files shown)


DirectXMan12:
This PR is quite old, and no longer compatible with the current websockify code.  If you are still interested, please update your code and reopen the PR.

lzap:
Sure, before I do that, can you review what I am trying to do? Do you agree with my patch?

DirectXMan12:
TBH, I've got some qualms about it.  Blanket catching of exceptions like this is not the best.  Do you have a specific exception you think should be caught?  I concur that something like "address in use" should probably not generate a full stack trace, but you should catch that specifically.

lzap:
My particular problem is that we call your application several times until a free port is found, all is erroring with exception trace which is then caught by ABRT tool:

```
  File "/home/lzap/work/foreman/extras/noVNC/websockify.py", line 5, in <module>
    websockify.websocketproxy.websockify_init()
  File "/home/lzap/work/foreman/extras/noVNC/websockify/websocketproxy.py", line 390, in websockify_init
    server.start_server()
  File "/home/lzap/work/foreman/extras/noVNC/websockify/websocket.py", line 834, in start_server
    lsock = self.socket(self.listen_host, self.listen_port, False, self.prefer_ipv6)
  File "/home/lzap/work/foreman/extras/noVNC/websockify/websocket.py", line 198, in socket
    sock.bind(addrs[0][4])
  File "/usr/lib64/python2.7/socket.py", line 224, in meth
    return getattr(self._sock,name)(*args)
socket.error: [Errno 98] Address already in use
```

Do you want me to create a patch that will only catch this particular exception? I was under impression that generic approach might work nicely, users still see errors and if they want details `--verbose` can be provided...

PR #135, opened 2014-08-08
Remove explicit numpy requirement from install_requires

holzman:
According to the code, numpy is optional, not required.
(It improves the efficiency of hybi, but there's a fallback).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,6 @@ setup(name=name,
 
       packages=['websockify'],
       include_package_data=True,
-      install_requires=['numpy'],
       zip_safe=False,
       entry_points={
         'console_scripts': [


DirectXMan12:
While it is possible to use websockify without numpy, it's better to use it with numpy, and the majority of users probably want numpy pulled in by default.

holzman:
I respectfully disagree. It's a huge dependency to bring in for little benefit.

kanaka:
@holzman do you have any quantitative numbers? For example, round trip time for a few thousand small messages from browser to websockify to server and back. And time to send a few hundred meg from browser to server.

While this only effects client to server traffic (due to masking), my memory is that I've seen a noticeable difference in latency with numpy vs raw python. For example, in noVNC when moving the mouse around screen rapidly, the server's mouse will lag further behind without numpy. However, more importantly, noVNC is not the only user of websockify and for application where they might be a lot of client to server traffic, the difference could be quite significant. 

You might be right, but I would trade extra disk usage in exchange for performance any day especially for websockify. If you can show quantitatively that the performance difference is negligible for numpy vs raw python, then removing numpy from the default would make more sense to me. Or propose a different way of doing this in python that has equal or better performance in all important uses cases and with a smaller dependencies, I think that would be great.

holzman:
Fair enough, I haven't done any quantitative measurement. I'll do so and propose the appropriate change.

- B

On Apr 28, 2015, at 5:50 PM, "Joel Martin" <notifications@github.com<mailto:notifications@github.com>> wrote:


@holzman<https://github.com/holzman> do you have any quantitative numbers? For example, round trip time for a few thousand small messages from browser to websockify to server and back. And time to send a few hundred meg from browser to server.

While this only effects client to server traffic (due to masking), my memory is that I've seen a noticeable difference in latency with numpy vs raw python. For example, in noVNC when moving the mouse around screen rapidly, the server's mouse will lag further behind without numpy. However, more importantly, noVNC is not the only user of websockify and for application where they might be a lot of client to server traffic, the difference could be quite significant.

You might be right, but I would trade extra disk usage in exchange for performance any day especially for websockify. If you can show quantitatively that the performance difference is negligible for numpy vs raw python, then removing numpy from the default would make more sense to me. Or propose a different way of doing this in python that has equal or better performance in all important uses cases and with a smaller dependencies, I think that would be great.

Issue #17, opened 2011-11-03
Support non-Base64 Encoded Strings


jmcintyre:
In other frameworks, it has been possible to send strings through the websocket api without first encoding them into base64. For users that want to pass non-binary data, this would be easier to get up and running. It would also fit well with many examples online that are not specifying a protocol in the constructor of WebSocket and use simple string in the send call. Please consider adding this option.

PR #137, opened 2014-09-15
Add "text" subprotocol, and make it the default.

kousu:
Websockify previously forced you to specify, "binary" or "base64"
when constructing a WebSocket, forcing you to either handle Blobs
or start all your processing in btoa().

This patch gives a mode to websockify to pass data unprocessed without
marking it with the 0x2 opcode that makes browsers interpret websocket
frames as Blobs. This lets client code that works against other WebSocket
implementations, such as AutobahnPython, to also work unchanged.

In code, the difference is, previously you would have to do:
```
var ws = new WebSocket("ws://localhost:8080", "binary")
ws.onmessage = function(e) {
  var F = new FileReader();
  var payload = null;
  F.readAsText(e.data);
  F.onloadend = function() {
    payload = F.result;
  }
}
```

or

```
var ws = new WebSocket("ws://localhost:8080", "base64")
ws.onmessage = function(e) {
  var payload = btoa(e.data);
}
```

Now you can just do

```
var ws = new WebSocket("ws://localhost:8080")
ws.onmessage = function(e) {
  var payload = e.data;
}
```

This patch still allows the old way to work (in particular, when pushing
lots of data, careful Blob processing should help performance.. I think.)

I believe that this patch fulfills the original request in #17. In that thread, @kanaka said "base64 needs to be the default" but that must have changed since then because I wasn't able to speak to websockify without specifying a protocol.

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -147,13 +147,16 @@ class WebSocketRequestHandler(SimpleHTTPRequestHandler):
     @staticmethod
     def encode_hybi(buf, opcode, base64=False):
         """ Encode a HyBi style WebSocket frame.
-        Optional opcode:
+        Opcode, defined by the HyBi spec:
             0x0 - continuation
-            0x1 - text frame (base64 encode buf)
-            0x2 - binary frame (use raw buf)
+            0x1 - text frame
+            0x2 - binary frame
             0x8 - connection close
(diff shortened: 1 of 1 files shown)


DirectXMan12:
closing due to noresponse

PR #146, opened 2014-12-06
Add Python 3 trove classifiers.

vaibhavsagar:
This makes <http://py3readiness.org> display the status of this package correctly :).

### DIFF
--- a/include/websock.js
+++ b/include/websock.js
@@ -262,7 +262,7 @@ function on(evt, handler) {
     eventHandlers[evt] = handler;
 }
 
-function init(protocols) {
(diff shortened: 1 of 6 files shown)


vaibhavsagar:
I did some rebasing, let me know if you'd like me to close this pull request and submit a new one instead.

DirectXMan12:
Yes, it would be great if you could submit a PR in a branch that is synced with noVNC master so we only have the one commi ("Add Python 3 trove classifiers").  Thanks!

vaibhavsagar:
Done, see #159.

PR #157, opened 2015-02-27
added --target-api option

james-portman:
I have added --target-api option to query tokens against a JSON Rest API

The API URL should be given by the --target-api option,
the token given will be appended to the API URL e.g. for a token of 1234abcd
http://localhost/theapi/ becomes http://localhost/theapi/1234abcd

ip and port of the VNC connection should be returned in JSON format from the API

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -11,7 +11,7 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import signal, socket, optparse, time, os, sys, subprocess, logging
+import signal, socket, optparse, time, os, sys, subprocess, logging, urllib2, json
 try:    from socketserver import ForkingMixIn
 except: from SocketServer import ForkingMixIn
 try:    from http.server import HTTPServer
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Hi!  Thanks  for the contribution.

A couple of commets:

1.  Can you please squash all the of the "cleanup" commits into one or two separate commits?
2. Since this is the second "option" for using tokens, what would be really great would be if we could have a plugin-type system that allowed noVNC to call out to a separate class/method to turn a token into a target pair.  Then, we could simply provide the current functionaly (reading from a file) and your functionality as example/built-in plugins (then you could do something like `--token-plugin websockify.token_loaders.FromFile --token-source file_name_here` or `--token-plugin ValidPythonImportPathHere --token-source "some argument be passed to the class on instantiation"`).

@kanaka : thoughts on number two?

james-portman:
Hi,

Sorry I only realised they weren't squashed afterward but thought it would
still be a good use case for you to see and talk about,

I do think it would be nice to split out the functionality to allow others
in future too yeah,

I will have a look when I get some spare time and do a new, and tidier pull
request

Thanks

DirectXMan12:
For future reference, you can just squash and `git push --force`, and GitHub will update the PR automatically

dhilipsiva:
@james-portman I sent you a pull request. There was a tiny bug in your pull request. Its working now.

james-portman:
Think I have managed to squash it by the way, I did try originally but the push --force made the difference, thanks.

dhilipsiva:
@james-portman Yeah, I just saw you squashed commit. I thought you may not find time and opened PR #158 and I closed after seeing your commit. Thanks for this PR.

DirectXMan12:
Closing in favor of #162.

PR #160, opened 2015-03-24
Added mechanism to ensure data send

vedmack:
Added mechanism to ensure data send in case that `websocket.bufferedAmunt > api.maxBufferedAmount`

Because sometimes when the `websocket.bufferedAmount > api.maxBufferedAmount` the `else` block of `flush` function was reached and eventually the websokets got closed, and that _mechanism with the `setInterval`_ is the only way that I managed to keep the connection alive and working.

Daniel.

### DIFF
--- a/include/websock.js
+++ b/include/websock.js
@@ -65,7 +65,8 @@ var api = {},         // Public API
         'error'   : function() {}
     },
 
-    test_mode = false;
+    test_mode = false,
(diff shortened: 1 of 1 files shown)


CendioOssman:
The buffer limit has been removed as of 40238b00b61e509260e3093b0b353d5dcef6c413.

PR #161, opened 2015-03-25
BUGFIX:  Websocket frame corruption on big-endian #150

### LINKED ISSUE #150: Websocket frame corruption on big-endian
I was try to connect to a VNC server using noVNC + Websockify on a SPARC system running Solaris. I noticed that X86 worked fine but the VNC session got terminated during handshake on SPARC. Further examination and on running the echo unit tests 'echo' and 'plain echo' revealed a possible byte order issue. I traced the issue down to the use of numpy to  apply the mask on the  data in the Websocket frame, which seems to assume that both will be in little-endian representation, which leads to a no-op on X86, but causes an invalid byte swap on big-endian ISAs like SPARC, some ARM , etc.

A very simple patch that resolved the problem for me is below.

Thanks,
Niall

--- websockify-0.5.1/websockify/websocket.py.orig       2015-01-13 14:59:26.3828
30765 -0700
+++ websockify-0.5.1/websockify/websocket.py    2015-01-13 14:58:26.088075181 -0
700
@@ -241,9 +241,9 @@
         if numpy:
             b = c = s2b('')
             if plen >= 4:
-                mask = numpy.frombuffer(buf, dtype=numpy.dtype('<u4'),
+                mask = numpy.frombuffer(buf, dtype=numpy.dtype('u4'),
                         offset=hlen, count=1)
-                data = numpy.frombuffer(buf, dtype=numpy.dtype('<u4'),
+                data = numpy.frombuffer(buf, dtype=numpy.dtype('u4'),
                         offset=pstart, count=int(plen / 4))
                 #b = numpy.bitwise_xor(data, mask).data
                 b = numpy.bitwise_xor(data, mask).tostring()

rafaelfolco:
This patch fixes noVNC endianess problem on handshake.
It affects noVNC sessions on Big Endian platforms.

Fixes #150

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -118,20 +118,24 @@ class WebSocketRequestHandler(SimpleHTTPRequestHandler):
         if numpy:
             b = c = s2b('')
             if plen >= 4:
-                mask = numpy.frombuffer(buf, dtype=numpy.dtype('<u4'),
-                        offset=hlen, count=1)
-                data = numpy.frombuffer(buf, dtype=numpy.dtype('<u4'),
-                        offset=pstart, count=int(plen / 4))
+                dtype=numpy.dtype('<u4')
+                if sys.byteorder == 'big':
+                    dtype = dtype.newbyteorder('>')
+                mask = numpy.frombuffer(buf, dtype, offset=hlen, count=1)
+                data = numpy.frombuffer(buf, dtype, offset=pstart,
+                        count=int(plen / 4))
                 #b = numpy.bitwise_xor(data, mask).data
                 b = numpy.bitwise_xor(data, mask).tostring()
 
             if plen % 4:
                 #self.msg("Partial unmask")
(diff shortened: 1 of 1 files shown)


DirectXMan12:
LGTM :+1:
@kanaka, @samhed , @astrand : any concerns?

samhed:
No, but those trailing whitespaces annoy me! :smile:

DirectXMan12:
@rafaelfolco : can you fix the whitespace, please?  Otherwise, this looks great.

kanaka:
Looks okay to me apart from whitespace. This is in the main data path and I'm always concerned about adding even small amounts of latency there, but one extra check for the big endian case shouldn't have any effect.

rafaelfolco:
commit d56a2e5a87e29be7e86aa5121efd74ac425b59c6 addresses the whitespace. Please let me know if any other concerns. Thanks.

samhed:
Could you just squash the two commits together? Just so we can keep our history cleaner. You can use git push --force after making the changes so you don't have to do a new PR.

PR #162, opened 2015-03-26
Introduce Token Plugins

DirectXMan12:
Token plugins provide a generic interface for transforming a token
into a `(host, port)` tuple.

The plugin name is specified using the '--token-plugin' option,
and may either be the name of a class from `websockify.token_plugins`,
or a fully qualified python path to the token plugin class (see below).

An optional plugin parameter can be specified using the '--token-source'
option (a value of `None` will be used if no '--token-source' option is
passed).

Token plugins should inherit from `websockify.token_plugins.BasePlugin`,
and should implement the `lookup(token)` method.  The value of the
'--token-source' option is available as `self.source`.

Several plugins are included by default.  The `ReadOnlyTokenFile`
and `TokenFile` plugins implement functionality from '--target-config'
(with the former only reading the file(s) once, and the latter reading
them every time).  The 'BaseTokenAPI' plugin fetches the value from
an API, returning the result of `process_result(response_object)`.
By default, `process_result` simply returns the text of the response,
but may be overriden.  The `JSONTokenAPI` does just this, returning
the 'host' and 'port' values from the response JSON object.

The old '--target-config' option is now deprecated, and maps to the
`TokenFile` plugin under the hood.

Closes #157


DirectXMan12:
@james-portman : does this work for you?

@kanaka : look good to you?

james-portman:
@DirectXMan12 , looks good to me, sorry I didn't have time to do this recently.

I did actually make another version which calls an external executable and uses the output, rather than being limited to just python, if that would be useful as an additional token plugin?
It would work in a similar way to the 'BaseTokenAPI' plugin but catch an executables output rather than a curl.

PR #163, opened 2015-03-27
don't make numpy a hard requirement

sdague:
the websockify code works without numpy. It will be slower, however it
only seems to be used if you are directly processing masked buffers in
python. However, the current code hard enforces numpy in setup.py, so
if you install the package you forcably get numpy.

In OpenStack we use the websockify proxy a lot of places where the
numpy paths aren't touched. Forcing numpy to be installed in all these
circumstances is somewhat overkill.

This would take it out of forced requirements.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,6 @@ setup(name=name,
 
       packages=['websockify'],
       include_package_data=True,
-      install_requires=['numpy'],
       zip_safe=False,
       entry_points={
         'console_scripts': [


DirectXMan12:
I am unclear as to why the numpy paths would be untouched.  RFC 6455 states that "a client MUST mask all frames that it sends to the server".  According to https://developer.mozilla.org/en-US/docs/WebSockets , RFC6455 support has been in place in most browsers for a while.

Could you please clarify where you use websockify that the numpy paths would not be touched?

sdague:
Ah, you are right, I didn't understand all the implications here. Thanks.

sdague:
I do wonder if that's the case why is numpy optional in the code?

DirectXMan12:
I suspect that it's because websockify was designed to be compatible with a wide range of environments and Python versions, so of which may not have had numpy support at all.  If someone wanted to run websockify in one of these environments, hypothetically they could clone the repository and run websockify from the repository, without having to actually patch any code (they just couldn't use `setup.py` or `pip install`).

Issue #166, opened 2015-04-04
Interrupted system call


bkanuka:
This is a bit of a cross-post because I'm not entirely sure where this bug lives.  [WebVirtMgr](https://github.com/retspen/webvirtmgr) uses websockify somewhat unsuccessfully. See [this bug](https://github.com/retspen/webvirtmgr/issues/503).

The actual exception however, is happening in [websocketproxy.py line 134](https://github.com/kanaka/websockify/blob/master/websockify/websocketproxy.py#L134). `ins, outs, excepts = select(rlist, wlist, [], 1)` is throwing a `select.error error: (4, 'Interrupted system call')`

If the `select` timeout is set to 0, things work but obviously at high CPU.

The bug linked above may actually be the fault of WebVirtMgr, but ideally websockify should be robust to whatever WebVirtMgr is doing.

bkanuka:
A reasonable reaction to an "Interrupted system call" is to try again. My patch just catches and continues (a little trickery to catch *just* "Interrupted system call")

haypo:
> A reasonable reaction to an "Interrupted system call" is to try again.

Yes. For select.select(), you may need to recompute the timeout. But in do_proxy() calls select() in a "while True" loop, so I'm not sure that the timeout matters. By the way, why select() is called with a timeout? It looks like active polling, whereas it should work without polling.

See also the PEP 475 which restarts interrupted syscalls in Python 3.5:
https://www.python.org/dev/peps/pep-0475/

bkanuka:
`select` is called with a timeout so that it blocks for *up to* one second. Without it, `select` never blocks and the encasing `while True` spins as fast as it can. I think the net effect is that the timeout doesn't really matter (as long as it's not 0).

Re: Python 3.x - it's the reason my patch wasn't merged. I'm not too familiar with Python 3 and especially not with code that's cross-compatible. Will have to look into this (fingers crossed for today).

Thank you for looking at this.

psych0d0g:
@bkanuka after i've applied this patch, WebVirtMgr isnt connecting to spice at all anymore, without it connects but disconnects after few seconds like the initial issue, something seems not quite right here i guess

bkanuka:
Whoa super good to know. Thanks! I'll look into it. Sorry this has totally
slipped away from me.

PR #167, opened 2015-04-05
Catch interrupted system call

bkanuka:
Patch for Issue #166

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -11,12 +11,12 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import signal, socket, optparse, time, os, sys, subprocess, logging
+import signal, socket, optparse, time, os, sys, subprocess, logging, errno
 try:    from socketserver import ForkingMixIn
 except: from SocketServer import ForkingMixIn
 try:    from http.server import HTTPServer
 except: from BaseHTTPServer import HTTPServer
-from select import select
+import select
 from websockify import websocket
 try:
     from urllib.parse import parse_qs, urlparse
@@ -131,7 +131,14 @@ Traffic Legend:
 
             if tqueue: wlist.append(target)
             if cqueue or c_pend: wlist.append(self.request)
-            ins, outs, excepts = select(rlist, wlist, [], 1)
+            try:
+                ins, outs, excepts = select.select(rlist, wlist, [], 1)
+            except select.error, err:
+                if err[0] != errno.EINTR:
+                    raise
+                else:
+                    continue
+                
             if excepts: raise Exception("Socket exception")
 
             if self.request in outs:


DirectXMan12:
In general, I think this looks good, but please fix the `except` clause.

bkanuka:
What Python version should I be targeting?

kanaka:
Earliest is python 2.4. I keep thinking that we'll be able to get away from this one, but every time we break it we get bug reports so people are still apparently using python 2.4 even today. You should also test it with a python 3.X version (probably python 3.4 if you have it but any python 3.X will probably suffice).

DirectXMan12:
@kanaka : I think, at some point, we should just abandon Python 2.4 support (and probably 2.5), and set a roadmap or policy for deprecating Python versions.

kanaka:
@DirectXMan12 definitely. Maybe we could put a deprecation notice in the code that if you want websockify to continue supporting python2.4 please go vote at this issue. We might get some pulse on how many people are actually still dependent on it. Or maybe refer them to a branch of an older version (although I would hate to get in the business of supporting a bunch of older versions for security, etc).

DirectXMan12:
@kanaka : We're probably coming up on a new release sometime soon (there's a couple things I want to get in first).  I'll add it to the release notes in *big* **bold** ***LETTERS***.  Is that ok?  I'm not aware of any supported platforms that use Python 2.4.  Normally I'm all for deprecation, but I think it's fine in this case just to drop support entirely.

kanaka:
Yes, that's fine.

astrand:
In ThinLinc, our base requirement is Python 2.4 or newer. (See https://www.cendio.com/resources/docs/tag/requirements_server.html). We are using websocket.py.

kanaka:
@astrand so just to be clear, you are wanting to not deprecate python 2.4 compatibility for the time being? Do you have any sense of when that would no longer be needed for your situation?

DirectXMan12:
@astrand : the three "example" Linux platforms listed for ThinLinc all have newer versions of Python.  Python 2.4 hasn't been in most distros for a while (Python 2.4 was in RHEL 5, which is 8 years old at this point, and thus is rapidly nearing EOL).

astrand:
Although we are mainly testing and recommending newer versions such as RHEL7, we still have customers running older versions. That's why our requirements are allowing this. As far as I know, we do not have any decision or information about when we will raise the Python requirement. I believe we still have customers running RHEL 5. For example, see this bug report, which is less than a year old: https://www.cendio.com/bugzilla/show_bug.cgi?id=5180 . So let it put it like this: Is there any real problem with continuing to support Python 2.4? Otherwise, my suggestion is to try to keep supporting that version at least until RHEL5 reaches (non-extended) EOL. According to http://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux, the "End of Production 3 Phase" is 31 March 2017.

DirectXMan12:
@astrand 

> Is there any real problem with continuing to support Python 2.4? 

there's a couple of odd work-arounds that are needed to maintain compatiblity with both Python 2.4 and Python 3.x.  For instance, the exception syntax used in Python 3.x (`except Exception as ex`) was backported to Python 2.6 (which is the version that ships with RHEL 6 and Ubuntu 10.04 LTS).  Otherwise, to capture the actual exception, you have to do something like this: https://github.com/kanaka/websockify/blob/master/websockify/websocket.py#L519.

This places a burden on the submitters (they have to know that code has to work with Python 2.4 and Python 3.3+ and test against that), and also somewhat on us (the reviewers).

> I believe we still have customers running RHEL 5. For example, see this bug report, which is less than a year old: https://www.cendio.com/bugzilla/show_bug.cgi?id=5180

Of course.  I'm no stranger to this myself.  However, at a certain point, generally new releases of the project move forward with new versions of Python, and the older versions simply continue to get bugfixes.  Does each version of ThinLinc have a corresponding version of websockify that it ships?  If so, do you have RHEL 5 customers running the latest version of ThinLinc?

We could consider doing a "legacy" branch that just gets appropriate fixes backported, although we don't really get a high number of feature requests against websockify, so I suspect that would be more effort than it's worth.

> Otherwise, my suggestion is to try to keep supporting that version at least until RHEL5 reaches (non-extended) EOL. According to http://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux, the "End of Production 3 Phase" is 31 March 2017.

This isn't a bad plan, although my opinion is that we should drop support sooner ;-)

DirectXMan12:
obsoleted by #175

astrand:
Sorry about the slow response, I'm on paternity leave. Just to follow up on some of the questions:

>"Does each version of ThinLinc have a corresponding version of websockify that it ships? If so, do you have RHEL 5 customers running the latest version of ThinLinc?"

We are shipping a single version of ThinLinc which works on all supported distributions. Yes, I believe we have customers running the latest ThinLinc on RHEL5. 

Creating a branch sounds fine with me as well.

PR #172, opened 2015-04-29
Auth Plugins / Plugin Cleanup

DirectXMan12:
This PR introduces auth hooks, with allow websockify users to easily add in arbitrary authentication without having to subclass the main classes of websockify.  Authentication plugins are passed the header object, as well as the target host and port.  When initialized, they are passed a single parameter, similarly to token plugins.  If authentication fails, auth plugins should raise an exception.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ other/node_modules
 .pydevproject
 target.cfg
 target.cfg.d
+.tox
+*.egg-info
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: python
+python:
+    - 2.6
+    - 2.7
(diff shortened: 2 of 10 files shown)


DirectXMan12:
@kanaka , @samhed , @aastrand : does this look good to you?

There were several older bugs and PRs that requested more specific forms of this feature, so I decided to generalize.

kanaka:
@DirectXMan12 I did a brief scan of the change and it looks good to me. Maybe a test for this?

DirectXMan12:
@kanaka : I'll write up a couple tests.

DirectXMan12:
Ok, I rewrote the tests to actually work, and enabled Travis along the way

PR #174, opened 2015-05-12
Introduce strict mode

### LINKED ISSUE #164: websockify accepts unmasked frames
**Cross-filed from https://bugs.launchpad.net/nova/+bug/1278342**

Currently, websockify will happily accept unmasked frames, simply logging a debug message when it gets one (see [websockify/websockify.py:240](https://github.com/kanaka/websockify/blob/master/websockify/websocket.py#L240)).

However, according to RFC 6455 (section 5.1),
> The server MUST close the connection upon receiving a frame that is not masked.

For compatibility reasons, we should continue to have an option to have the current behavior.  However, we should also probably introduce a "strict mode" (on by default in new releases) which follows the RFC and rejects unmasked frames.

@sdague, @kanaka , @samhed , @astrand : thoughts on the proposed solution?

DirectXMan12:
This commit introduces strict mode, which is on by default.  Currently
strict mode only enforces client-to-server frame masking.  However,
in the future, it might enforce other parts of the RFC as well.

Closes #164

### DIFF
--- a/tests/test_websocket.py
+++ b/tests/test_websocket.py
@@ -343,3 +343,88 @@ class WebSocketServerTestCase(unittest.TestCase):
                                             socket.TCP_KEEPIDLE), keepidle)
         self.assertNotEqual(sock.getsockopt(socket.SOL_TCP,
                                             socket.TCP_KEEPINTVL), keepintvl)
+
+
+class HyBiEncodeDecodeTestCase(unittest.TestCase):
+    def test_decode_hybi_text(self):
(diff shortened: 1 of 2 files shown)


DirectXMan12:
@kanaka , @samhed : look good for what we talked about in #164 ?

DirectXMan12:
(I'll add tests once the test refactor in #172 lands)

kanaka:
Looks good to me.

samhed:
Great!

PR #178, opened 2015-06-01
Avoid UnboundLocalError for exc

no2a:
Previously exc was set only when an exception occurs in the inner
try (the one contains call to self.poll), so an error in the other
part causes an UnboundLocalError.

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -1047,6 +1047,7 @@ class WebSocketServer(object):
                         self.msg("In exit")
                         break
                     except Exception:
+                        _, exc, _ = sys.exc_info()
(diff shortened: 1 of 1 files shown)


kanaka:
This change is currently not portable across the versions of python that we support (2.4-3.X). See: http://stackoverflow.com/questions/12682558/how-to-write-an-exception-catching-code-works-in-python2-4-to-python3

Can you update it to be portable (as described above) and update the pull request? Thanks

no2a:
Changed to get exc using sys.exc_info() as the other part of the file.

DirectXMan12:
Can you squash this into one commit please?  Then it should be all set to merge.  Thanks!

PR #190, opened 2015-08-11
Switch to using ssl.create_default_context 

devicenull:
The default settings that you get with ssl.wrap_socket are very insecure.  They support SSLv2 and SSLv3, and do not provide any ciphers supporting forward secrecy.

Prior to these changes, a websockify server gets a 'C' grade on the ssllabs test.  It's also vulnerable to POODLE.

After these changes, it now gets an 'A' grade, as well as supporting forward secrecy.

The downside here is that this would drop support for anything lower then Python 2.7.9.

I'm relying on a ssl module function here to pick secure defaults, so this shouldn't require chasing a moving target: https://docs.python.org/2.7/library/ssl.html#ssl.create_default_context

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -815,11 +815,12 @@ class WebSocketServer(object):
                                   % self.cert)
             retsock = None
             try:
-                retsock = ssl.wrap_socket(
+                context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+                context.load_cert_chain(self.cert, self.key)
+
+                retsock = context.wrap_socket(
                         sock,
-                        server_side=True,
-                        certfile=self.cert,
-                        keyfile=self.key)
+                        server_side=True)
             except ssl.SSLError:
                 _, x, _ = sys.exc_info()
                 if x.args[0] == ssl.SSL_ERROR_EOF:


devicenull:
I don't understand what that test is meant to actually test.  This line in particular doesn't make sense:

https://github.com/kanaka/websockify/blob/master/tests/test_websocket.py#L94

Why would the key be set to a tmpdir?

kanaka:
@devicenull It looks like this change by @hopem changed the key setting from "./" to a tmpdir: https://github.com/kanaka/websockify/commit/32c1abd5d9643296e0d30abe2b2ccde324d5abcc

And the original change that added the "./" is this: https://github.com/kanaka/websockify/commit/081046b6cd04fd2cc84488bf5862d31a016fd0ad

It's been two years, but I think it was just an easy default that got the basic tests to run. However, it looks like the reason the [test test_do_handshake_ssl_error_eof_raises_close_error](https://github.com/kanaka/websockify/blob/master/tests/test_websocket.py#L253) is failing is because it is still monkey patching of ssl.wrap_socket rather than the context.wrap_socket. Not sure it will be feasible to patch that so test may need to be refactored.

DirectXMan12:
However, we did promise that we'd support older versions of Python.  Can you add in some fallback code?  We've indicated in the past that we'll support Python 2.4 until RHEL 5 is at EOL (see https://github.com/kanaka/websockify/pull/167).

CendioOssman:
This got fixed through #308.

PR #192, opened 2015-08-22
Fix bug in JSONTokenApi plugin

grz0:
Fixes:
>handler exception: 'method' object is not subscriptable
exception
  File "/root/websockify/websockify/websocket.py", line 910, in top_new_client
    client = self.do_handshake(startsock, address)
  File "/root/websockify/websockify/websocket.py", line 845, in do_handshake
    self.RequestHandlerClass(retsock, address, self)
  File "/root/websockify/websockify/websocket.py", line 114, in __init__
    SimpleHTTPRequestHandler.__init__(self, req, addr, server)
  File "/usr/lib/python3.4/socketserver.py", line 673, in __init__
    self.handle()
Fixes: 
 File "/root/websockify/websockify/websocket.py", line 570, in handle
    SimpleHTTPRequestHandler.handle(self)
  File "/usr/lib/python3.4/http/server.py", line 398, in handle
    self.handle_one_request()
  File "/usr/lib/python3.4/http/server.py", line 386, in handle_one_request
    method()
  File "/root/websockify/websockify/websocket.py", line 536, in do_GET
    if not self.handle_websocket():
  File "/root/websockify/websockify/websocket.py", line 524, in handle_websocket
    self.new_websocket_client()
  File "/root/websockify/websockify/websocketproxy.py", line 48, in new_websocket_client
    (self.server.target_host, self.server.target_port) = self.get_target(self.server.token_plugin, self.path)
  File "/root/websockify/websockify/websocketproxy.py", line 103, in get_target
    result_pair = target_plugin.lookup(token)
  File "/root/websockify/websockify/token_plugins.py", line 72, in lookup
    return self.process_result(resp)
  File "/root/websockify/websockify/token_plugins.py", line 82, in process_result
    return (resp.json['host'], resp.json['port'])
TypeError: 'method' object is not subscriptable

### DIFF
--- a/websockify/token_plugins.py
+++ b/websockify/token_plugins.py
@@ -79,4 +79,4 @@ class JSONTokenApi(BaseTokenAPI):
     # should go
 
     def process_result(self, resp):
-        return (resp.json['host'], resp.json['port'])
+        return (str(resp.json()['host']), str(resp.json()['port']))


websockify/token_plugins.py
```diff
@@ -79,4 +79,4 @@ class JSONTokenApi(BaseTokenAPI):
     # should go
 
     def process_result(self, resp):
-        return (resp.json['host'], resp.json['port'])
+        return (str(resp.json()['host']), str(resp.json()['port']))
```
DirectXMan12: I don't think the calls to `str()` are necessary here.

DirectXMan12:
In general, LGTM :+1:.  Just one inline comment

websockify/token_plugins.py
peprasetya: On my test, I have error message:
handler exception: Int or String expected

When I change the line to:
return (str(resp.json['host']), str(resp.json['port']))

it solve the error. So on my case the str() is needed.
And the resp.json()[] is not work, so it should use just resp.json[].

 My python is version 2.6.6

PR #194, opened 2015-08-25
Rework Auth Plugins to Support HTTP Auth

DirectXMan12:
This commit reworks auth plugins slightly to enable
support for HTTP authentication.  By raising an
AuthenticationError, auth plugins can now return
HTTP responses to the upgrade request (such as 401).

Related to kanaka/noVNC#522

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -18,6 +18,7 @@ try:    from http.server import HTTPServer
 except: from BaseHTTPServer import HTTPServer
 import select
 from websockify import websocket
+from websockify import auth_plugins as auth
 try:
     from urllib.parse import parse_qs, urlparse
 except:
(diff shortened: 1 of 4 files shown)


DirectXMan12:
@kanaka @samhed thoughts?

websockify/websocketproxy.py
```diff
@@ -37,20 +38,34 @@ class ProxyRequestHandler(websocket.WebSocketRequestHandler):
...
+
+        if self.server.auth_plugin:
+            try:
+                self.server.auth_plugin.authenticate(
+                    headers=self.headers, target_host=self.server.target_host,
+                    target_port=self.server.target_port)
+            except auth.AuthenticationError:
+                ex = sys.exc_info()[1]
+                self.send_auth_error(ex)
+                raise
```
kanaka: Seems like auth should be before token plugin. Reduces the ability for non-authorized connections to do a denial of service to the system by inducing disk or DB activity due to token authorization.

kanaka:
One inline comment but otherwise looks fine to me.

websockify/websocketproxy.py
DirectXMan12: The issue is that currently, the `authenticate` method receives the decode host and port (that way you could say `person X is only authorized to connect to host/port Y`), so changing it would break backwards compatibility.  Hmm...

PR #198, opened 2015-10-05
Clarified SSL section in README

halfgaar:


kanaka:
@halfgaar that's really helpful. Thanks!

PR #201, opened 2015-10-06
Verify username/password with BasicAuth plugin

quinox:
```python
pull_request_description = "no comment"
```


DirectXMan12:
Heh, whoops, good catch.  Not sure how that slipped in there...
Man, I must have been *really* tired that day :-(

PR #202, opened 2015-10-06
Pythonize code

quinox:
Just a few nitpicks, feel free to reject this PR.


kanaka:
@quinox yeah, I'm not a huge fan of using try/catch for regular flow control. I know it's common practice, although I'm not sure I would consider it more "pythonic" in this case (especially for the token processing since the exception is often the common case).

Also, the use of urlparse that way isn't supported in python2.4 which is one of the environments that we support. Unfortunately, travis-ci doesn't directly support python2.4 so that's why it passed the tests, but python2.4 is something that we'll probably be supporting for a while longer.

PR #214, opened 2015-12-02
Fix log files when daemonizing

miguelxpn:
Save log file descriptor so it's not closed when calling daemonize()

I apologize for not catching this earlier.


DirectXMan12:
No problem.  I'm thinking about deprecating the `--daemon` option anyway.  Anyway, LGTM :+1:

Issue #193, opened 2015-08-25
multiprocessing exception: No Child processes


jeffforbes:
I recently modified the logging in a recent version of websocket.py so that the all of the loggeded information would go to a log file. I am not sure if my changes have caused a recurrence of the multiprocessing exception that was fixed in 2013. This error occurs on a RHEL-5 system with python2.4

Below in the error trace :

2015-08-20 03:00:58,182 exception
Traceback (most recent call last):
File "/opt/websockify/websocket.py", line 977, in start_server
child_count = len(multiprocessing.active_children())
File "/usr/lib64/python2.4/site-packages/multiprocessing/process.py", line 48, in active_children
_cleanup()
File "/usr/lib64/python2.4/site-packages/multiprocessing/process.py", line 58, in _cleanup
if p._popen.poll() is not None:
File "/usr/lib64/python2.4/site-packages/multiprocessing/forking.py", line 106, in poll
pid, sts = os.waitpid(self.pid, flag)
OSError: [Errno 10] No child processes
2015-08-20 03:00:58,182 handler exception: [Errno 10] No child processes

The error only shows up on a busy site which is spawning many child processes.
When this error occurs, the parent process CPU eventually goes to 100% and the log file get filled with these messages. if verbose is off, then only the "OSError: [Errno 10] No child processes" is recorded.

Any ideas what may be causing this exception?

I have modified the code for logging to go to a file.
I changed the reference to self.log_message to  self.msg so that the location of the logged messages
would be consistent. I then set up a file handler to replace the streamhandler with a filehandler. I then added a new command line argument to turn on/off the logging to a file. I don't think that these logging changes would affect the multiprocessing module. I can put up a diff if it would be helpful.

DirectXMan12:
X-Post from kanaka/noVNC#523.  Can you post a diff of the new code please?

jeffforbes:
The first file is  the one that I am testing and the second is current one off of the github source

#diff /opt/websockify/websocketproxy.py ./websocketproxy.py
14c14
< import signal, socket, optparse, time, os, sys, subprocess, logging
---
> import signal, socket, optparse, time, os, sys, subprocess, logging, errno
19c19
< from select import select
---
> import select
49a50,54
>         if self.server.auth_plugin:
>             self.server.auth_plugin.authenticate(
>                 headers=self.headers, target_host=self.server.target_host,
>                 target_port=self.server.target_port)
> 
61c66
<         self.msg(msg)
---
[...]

PR #219, opened 2016-01-05
Remove additional signal calls in websockify that causes novnc to hang.

noah8713:
Openstack nova novnc-proxy services uses websockify to provide support
for nova vms using novnc proxy. At present, novnc hangs every couple of
weeks. It only resumes post restart of the novnc-proxy which is not
good. Hence, this code in websockify is updated to get rid of additional
signal calls to avoid novnc going in hang state even though process is
running. Basically, we are getting rid of existing msg and vmsg calls in
the websocket.py. This is kind of quick fix but we will need an
additional way of figuring out the logging to make it easy to trace in
case of any further failures in future.

closes bug: https://github.com/kanaka/noVNC/issues/556

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -895,11 +895,14 @@ class WebSocketServer(object):
         raise self.Terminate()
 
     def multiprocessing_SIGCHLD(self, sig, stack):
-        self.vmsg('Reaping zombies, active child count is %s', len(multiprocessing.active_children()))
+        # TODO: figure out a way to actually log this information without
+        #       calling `log` in the signal handlers
+        multiprocessing.active_children()
 
     def fallback_SIGCHLD(self, sig, stack):
         # Reap zombies when using os.fork() (python 2.4)
-        self.vmsg("Got SIGCHLD, reaping zombies")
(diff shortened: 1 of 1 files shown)


noah8713:
@kanaka kindly see if this works. Since the description is long, I did commit thrice as the message description was not aligning proper spaces.

DirectXMan12:
@kanaka, this look good to you?  I'd like to figure out a way to log the messages somehow, but I think for the time being this fix is fine with me.

@noah8713 I'd prefer if you collapsed the commits together, but I can do that myself when I merge.

noah8713:
@DirectXMan12 sounds good and thanks since you guys have admin access. :)

jeffforbes:
This may also fix the issue I submittedL
multiprocessing exception: No Child processes #193

noah8713:
@kanaka : Can you please help for merging the same?

noah8713:
Tried to squash and merge into a single commit for my change.

noah8713:
@DirectXMan12 and @kanaka any updates on this? :(

kanaka:
LGTM

noah8713:
Sweet. :+1:

PR #222, opened 2016-01-12
Pull numpy from requirements

cudadog:
Love that numpy isn't required for websockify, bummed that setup.py requires it.


DirectXMan12:
See the discussion on #212 please.

PR #226, opened 2016-02-04
websockify does not react to SIGTERM

### LINKED ISSUE #138: websockify does not react to SIGTERM
Running websockify as normal (e.g. `websockify 6080 127.0.0.1:5900`), and then sending a SIGTERM to the process (e.g. `kill <websockify-pid>`) does nothing if there is a client connected. I would expect it to disconnect any clients and terminate itself. Instead, I seem to have to `kill -9 <websockify-master-pid>` to get this behavior 

If you SIGTERM the child process, it will kill that one client and the master will continue running, but this part is expected behavior as far as I am concerned

abhishekkekane:
Child processes are not terminated when parent process of websockify
is killed.

In this fix made provision to terminate all child processes when
parent process is killed.

Fixes #138

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -1079,6 +1079,16 @@ class WebSocketServer(object):
 
                     except (self.Terminate, SystemExit, KeyboardInterrupt):
                         self.msg("In exit")
+                        # terminate all child processes
+                        if multiprocessing and not self.run_once:
+                            childs = multiprocessing.active_children()
+
+                            for child in childs:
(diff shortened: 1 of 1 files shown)


DirectXMan12:
I've made a couple of tweaks (I removed an unnecessary PID check, since active_children is supposed to only return children of the current process, and make some tweaks to the commit message).  In general LGTM :+1:

DirectXMan12:
closed with 4e6938e

PR #231, opened 2016-03-28
Unquote url path for log_message method.

tolbkni:
Replace %xx escapes by their single-character equivalent explicitly,
what BaseHTTPServer (Py2) or http.server (Py3) does not do,
otherwise log_message will deal with %xx escapes as %-formatting.
And will raise TypeError exception: not enough arguments for format
string.

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -35,6 +35,8 @@ try:    from io import StringIO
 except: from cStringIO import StringIO
 try:    from http.server import SimpleHTTPRequestHandler
 except: from SimpleHTTPServer import SimpleHTTPRequestHandler
+try:    from urllib.parse import unquote
+except: from urllib import unquote
 
 # python 2.6 differences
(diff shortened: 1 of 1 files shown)


CendioOssman:
I don't see how that issue is possible. Can you give an example?

tolbkni:
Hi @CendioOssman, for example, if I visit `/websockify?token=5e64d1f323784cbb94892d17d1b5920f&title=%E4%B8%AD%E6%96%87%E5%90%8D` or `/websockify?token=5e64d1f323784cbb94892d%2017d1b5920f`, it will raise TypeError exception: not enough arguments for format string.

CendioOssman:
That works perfectly here. But I see the problem in your upstream report; you're using an ancient version of websockify. This issue got fixed three years ago in 6c1a2e9032e72cfdf4a2cb62a41b1f233e9e9014.

So please upgrade and everything should be fine.

PR #239, opened 2016-06-30
Switch test dependency to mox3

javacruft:
mox is pretty much unmaintained these days, however the OpenStack
project are actively maintaining mox3 (a Python 3 compatibile fork
with some other improvements).

websockify seems quite happy to use mox3 instead, so switch the
test dependency and associated imports to use mox3.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,2 +1,2 @@
-mox
+mox3
 nose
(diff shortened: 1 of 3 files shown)


DirectXMan12:
LGTM :+1:

DirectXMan12:
Thanks!

PR #244, opened 2016-08-06
split header and data functions

totaam:
Main benefits:
- the functions can be re-used and tested more easily
- it is possible to send the data without first concatenating the header and data together, which can make quite a bit of difference for performance - but note that this depends on a number of factors: size of the payload, memory pressure, network speed, naggle enabled or not, etc.. 
  The threshold is set arbitrarily at 4KB, you could turn it into a constant and if this is set high enough you would not see any difference with the current version of the code, except it will perform better with large packets.

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -327,21 +327,7 @@ class WebSocketRequestHandler(SimpleHTTPRequestHandler):
                     opcode = 1
                 else:
                     opcode = 2
-                encbufs = self.send_hybi(buf, opcode, base64=self.base64, record=self.rec)
-
-        while self.send_parts:
-            # Send pending frames
-            buf = self.send_parts.pop(0)
-            sent = self.request.send(buf)
-
-            if sent == len(buf):
-                self.print_traffic("<")
-            else:
-                self.print_traffic("<.")
-                self.send_parts.insert(0, buf[sent:])
-                break
-
-        return len(self.send_parts)
+                self.send_hybi(buf, opcode, base64=self.base64, record=self.rec)
 
     def recv_frames(self):
         """ Receive and decode WebSocket frames.
@@ -494,7 +480,6 @@ class WebSocketRequestHandler(SimpleHTTPRequestHandler):
             # Indicate to server that a Websocket upgrade was done
             self.server.ws_connection = True
             # Initialize per client settings
-            self.send_parts = []
             self.recv_part  = None
             self.start_time = int(time.time()*1000)
 
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -50,10 +50,15 @@ except:
         return struct.unpack(fmt, slice)
 
 # Degraded functionality if these imports are missing
-for mod, msg in [('numpy', 'HyBi protocol will be slower'),
-                 ('ssl', 'TLS/SSL/wss is disabled'),
-                 ('multiprocessing', 'Multi-Processing is disabled'),
-                 ('resource', 'daemonizing is disabled')]:
+OPTIONAL_FEATURES = [
+                     ('numpy', 'HyBi protocol will be slower'),
+                     ('ssl', 'TLS/SSL/wss is disabled'),
+                     ('multiprocessing', 'Multi-Processing is disabled'),
+                    ]
+if not sys.platform.startswith("win"):
+    #resource module is not available on Windows
+    OPTIONAL_FEATURES.append(('resource', 'daemonizing is disabled'))
+for mod, msg in OPTIONAL_FEATURES:
     try:
         globals()[mod] = __import__(mod)
     except ImportError:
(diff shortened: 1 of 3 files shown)


websockify/websocket.py
```diff
@@ -310,17 +323,8 @@ def send_frames(self, bufs=None):
...
-                    encbuf, lenhead, lentail = self.encode_hybi(buf, opcode=2, base64=False)
-
-                if self.rec:
-                    self.rec.write("%s,\n" %
-                            repr("{%s{" % tdelta
-                                + encbuf[lenhead:len(encbuf)-lentail]))
-
-                self.send_parts.append(encbuf)
+                opcode = 2-int(self.base64)     #based64: opcode=1, binary: opcode=2
+                encbufs = self.send_hybi(buf, opcode, base64=self.base64, record=self.rec)
```
kanaka: You dropped the recording functionality. Also, the condition opcode calculation, while more concise, obfuscates the actual logic that is happening; the original is preferable to me.

kanaka:
I'm okay with this change (apart from my inline comment). Have you tested with python 2.4 and python 3.X?

websockify/websocket.py
kanaka: Oh, nevermind about the record functionality, I'm just blind I guess.

kanaka:
@totaam out of curiosity, how has the html5 interface in xpra been working out? I assume this was a performance enhancement discovered as part of using websockify in xpra?

totaam:
> Have you tested with python 2.4 and python 3.X?
Can't remember about 2.4, will re-check and get back to you.

> apart from my inline comment
The opcode thing, sure. Back to a simple "if-else" instead?

> how has the html5 interface in xpra been working out?
Great. In order to support some of the best features of xpra (sound forwarding, video encodings, etc), we've had to make a lot of drastic changes to the html5 code recently.
Many of those changes are due to land for the next release... overdue by 2 months already... Real soon now I am told.

The network performance issues we found turned out to be in the xpra html5 network code, but looking into websockify was useful anyway. I was going to quantify the benefits of this patch, but moved on to other things.

kanaka:
Yeah, simple if-else is preferable to me.

It's been a couple years since I've played with xpra, I'll have to try it again soon.

totaam:
Done.
Also changed the header and data merge code back to using "+" as this works on all versions of Python.
Tested on CentOS5 (Python 2.4) and Fedora 24 (Python 2.7.12 and 3.5.1)

websockify/websocket.py
```diff
@@ -311,16 +324,10 @@ def send_frames(self, bufs=None):
...
-                    encbuf, lenhead, lentail = self.encode_hybi(buf, opcode=2, base64=False)
-
-                if self.rec:
-                    self.rec.write("%s,\n" %
-                            repr("{%s{" % tdelta
-                                + encbuf[lenhead:len(encbuf)-lentail]))
-
-                self.send_parts.append(encbuf)
+                    opcode = 2
+                encbufs = self.send_hybi(buf, opcode, base64=self.base64, record=self.rec)
```
DirectXMan12: `send_hybi` doesn't look like it returns anything.  Also, doesn't this make the `whiel self.send_parts` below redundant?

DirectXMan12:
LGTM modulo the comment above.  Will merge once that's addressed.

samhed:
> but note that this depends on a number of factors: size of the payload, memory pressure, network speed, naggle enabled or not, etc..

samhed:
> but note that this depends on a number of factors: size of the payload, memory pressure, network speed, naggle enabled or not, etc.. 

Nagle was disabled for websockify in f23780eb how does this affect things here?

totaam:
@DirectXMan12 right you are, the send_parts code wasn't used at all either.
The commit above gets rid of it. (removes 17 lines of unused code)

@samhed this shouldn't make much of a difference even with naggle disabled: the packet join size of 4K is high enough that you won't be sending just the packet header alone. Even then, it may or may not cost an extra TCP packet for payload sizes above 4K. (depending on the OS network stack, send queue levels, etc..)
If you're concerned about the network traffic vs cpu trade-off of this change, we can bump the threshold to 32KB or higher. (at which point the potential network cost increase becomes completely negligible, whereas the extra memory copy becomes an unnecessary burden on the CPU)

What some transports do is to disable naggle when sending multiple chunks and re-enable it immediately after. (I don't think it is worth the hassle - could be worth a try)

websockify/websocket.py
```diff
@@ -164,21 +169,34 @@ def encode_hybi(buf, opcode, base64=False):
...
+        return (header, buf)
 
-        #self.msg("Encoded: %s", repr(header + buf))
+    def send_hybi(self, buf, opcode, base64=False, record=None):
+        """ Send a HyBi style WebSocket frame. """
+        header, buf = self.encode_hybi(buf, opcode, base64)
+        if record:
+            record.write("%s,\n" % repr("{%s{" % tdelta + encbufs[1]))
+        if len(buf)<=4096:
+            self.request.send(header+buf)
```
DirectXMan12: I didn't catch this earlier, but it looks live we've lost detection of partial sends here, which could be bad.

totaam: How so?
I don't see it in the quoted lines.

DirectXMan12: because socket.send isn't guaranteed to have sent all the requested data, and it returns however much it sends.  Previously, the related methods here would return the result of self.request.send, and we would check to make sure the entirety of the buffer was sent.  If not, we'd remember the rest, and send it later.  This patch removes that functionality entirely (it can be seen here in the fact that we don't do anything with the result of `self.request.send`).

CendioOssman:
I'm a bit sceptical about this. Can you share some tests that show the performance issue/improvement?

PR #247, opened 2016-09-16
Split out generic parts

ossman:
Right now websocketserver.py is a bit of a kitchen sink. It has ssl, threading, select, logging and lots of other things. As such it is often not a good fit for integrating in to other projects. This PR tries to remedy that by separating the truly generic parts from the bonus features.

github doesn't show a nice diff here, but git with a bunch of -C and -M will. The basic principle is that websocketserver.py is split in to three parts:
- websocket.py: Raw WebSocket protocol handling, modelled after SSLSocket
- websocketserver.py: Simple WebSocket request handler/server in the same principle as Python's standard SocketServer
- websockifyserver.py: The new base for all the extra features

The idea is that the basic parts should be at a state where they could be merged with upstream Python.

I've noted this as a request for comments as it still need doc strings and such, but I wanted to first make sure that it isn't any major objections to this kind of split. It would make it easier for us that want to integrate your WebSocket handling elsewhere. :)

This PR removes the features on non-strict protocol parsing, and data recording. If those are really important then I can look at re-introducing them. It also adds stubs for WebSocket client mode, but no complete code yet.


kanaka:
I think the general direction is sound.

Can you provide an example diff command(s) that properly resolves differences. I played around with -C and -M but I didn't get anything particularly helpful.

Data recording is pretty important because it allows us to capture traces that we can replay later to test noVNC. When you say you removed non-strict protocol parsing, what are you referring to?

When I tried to test it just now I got:

```
./run 6080 localhost:5901
...
TypeError: __init__() got an unexpected keyword argument 'record'
```

Please make sure that you manually test that http, https, flash policy and websockets all work in the same instance of websockify. Alternately you could implement the test stubs for that list so that they actually do something :-)

Thanks

ossman:
> I think the general direction is sound.
> 
> Can you provide an example diff command(s) that properly resolves differences. I played around with -C and -M but I didn't get anything particularly helpful.
> 

This seems to do the trick:

```
git diff -C -C -M 72ce5c1ae8abbcd3f8bb5f4e93915f61a1047704
```

> Data recording is pretty important because it allows us to capture traces that we can replay later to test noVNC.

Alright. I'll have a look then.

> When you say you removed non-strict protocol parsing, what are you referring to?

There was a 'strict' parameter that made the parser overlook some requirements of the protocol. I didn't include that when moving things around so this code will kick out a peer that doesn't properly adhere to the protocol.

> 
> Please make sure that you manually test that http, https, flash policy and websockets all work in the same instance of websockify. Alternately you could implement the test stubs for that list so that they actually do something :-)
> 

Sorry. I was a bit rushed to get this up so I only ran the tests. :)

ossman:
Code is now at a more properly tested state. So I just need to sort out documentation and this should be good to go.

ossman:
It should be done now. Please review away. :)

ossman:
Updated commits to specify HTTP/1.1 (which is actually required by the spec.).

ossman:
Good to be merged? :)

ossman:
Ping?

samhed:
I don't feel qualified to merge this, @kanaka or @DirectXMan12 PTAL

CendioOssman:
So I now have the power to merge this via this account. :)

But I'd like to hear from @kanaka and @DirectXMan12 first. Any objections to this code? Do you need more time?

DirectXMan12:
Sorry, I've been super busy.  I'll look at it tomorrow (today at this point).

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-                                self.msg('listener exit due to --idle-timeout %s'
-                                            % self.idle_timeout)
-                                break
+    def _decode_hybi(self, buf):
+        """ Decode HyBi style WebSocket packets.
+        Returns:
+            {'fin'          : boolean,
+             'opcode'       : number,
+             'masked'       : boolean,
+             'length'       : encoded_length,
```
DirectXMan12: might want to name this `payload_length` so as to not confuse it with the previous definition of length

DirectXMan12: For that matter, why are we using a dict and not a named tuple or class with `__slots__` set here?  It's probably more efficient.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
+                if code is None:
+                    self.close_code = 1005
+                    self.close_reason = "No close status code specified by peer"
+                else:
+                    self.close_code = code
+                    if reason is not None:
+                        self.close_reason = reason
+
+                self.shutdown(code, reason)
+                return ''.encode("ascii")
```
DirectXMan12: again, maybe return `None` here?

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
+        # time to end things
+        if self._received_close and self._sent_close:
+            self._close()
+
+    def _send(self, data):
+        # Queues data and attempts to send it
+        self._send_buffer += data
+        self._flush()
+
+    def _send_str(self, string):
```
DirectXMan12: perhaps name it something besides `_send_str` if the doc string says it doesn't send ;-)

(`_queue_str`, for instance?)

websockify/websocketserver.py
```diff
@@ -0,0 +1,872 @@
...
+        globals()[mod] = None
+        print("WARNING: no '%s' module, %s" % (mod, msg))
+
+if multiprocessing and sys.platform == 'win32':
+    # make sockets pickle-able/inheritable
+    import multiprocessing.reduction
+
+from websockify.websocket import WebSocket, WebSocketWantReadError, WebSocketWantWriteError
+
+class Base64WebSocket(WebSocket):
```
DirectXMan12: I think we can probably drop this -- we decided a while ago to drop base64 support in noVNC, so I'm not sure we need base64 support in Websockify any longer

websockify/websocketserver.py
```diff
@@ -0,0 +1,872 @@
...
+
+        if bufs:
+            for buf in bufs:
+                if self.rec:
+                    self.rec.write("%s,\n" % repr("{%s{" % tdelta + buf))
+                self.send_parts.append(buf)
+
+        # Flush any previously queued data
+        try:
+            self.request.sendmsg('')
```
DirectXMan12: we should just have a `flush` method

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-    s2b = lambda s: s      # No-op
-    s2a = lambda s: [ord(c) for c in s]
-try:    from io import StringIO
-except: from cStringIO import StringIO
-try:    from http.server import SimpleHTTPRequestHandler
-except: from SimpleHTTPServer import SimpleHTTPRequestHandler
+try:
+    import ssl
+except ImportError:
+    ssl = None
```
ossman: We're still on 2.4 at Cendio, but that's partly because we haven't had a good excuse to raise the requirements. I can put in a separate commit that removes any fixes for pre-2.6.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-    * run_once: Handle a single request
-    * handler_id: A sequence number for this connection, appended to record filename
+    class WebSocketWantReadError(OSError):
+        def __init__(self):
+            OSError.__init__(self, errno.EWOULDBLOCK)
+    class WebSocketWantWriteError(OSError):
+        def __init__(self):
+            OSError.__init__(self, errno.EWOULDBLOCK)
+
+class WebSocket:
```
ossman: Fixed.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-        elif payload_len > 125 and payload_len < 65536:
-            header = pack('>BBH', b1, 126, payload_len)
-        elif payload_len >= 65536:
-            header = pack('>BBQ', b1, 127, payload_len)
+        self.socket = None
 
-        #self.msg("Encoded: %s", repr(header + buf))
+    def __getattr__(self, name):
+        # These methods are just redirected to the underlying socket
+        if name in ["fileno",
```
ossman: `SSLSocket` uses internal, undocumented API:s to be able to inherit from `socket`. That may be fine for something part of the Python library, but I don't think we should do that. Not until we're part of the library at least. :)

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
+                    "setblocking"]:
+            assert self.socket is not None
+            return getattr(self.socket, name)
+        else:
+            raise AttributeError("%s instance has no attribute '%s'" %
+                                 (self.__class__.__name__, name))
 
-        return header + buf, len(header), 0
+    def connect(self, uri, origin=None, protocols=[]):
+        """Establishes a new connection to a WebSocket server.
```
ossman: Not for websockify, no. But I figured if this was going to be a complete Websocket object, then being able to act as a client would seem natural.

Might be useful for testing if nothing else.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-                else:
-                    encbuf, lenhead, lentail = self.encode_hybi(buf, opcode=2, base64=False)
+            (request, self._recv_buffer) = self._recv_buffer.split('\r\n'.encode("ascii"), 1)
+            request = request.decode("latin-1")
 
-                if self.rec:
-                    self.rec.write("%s,\n" %
-                            repr("{%s{" % tdelta
-                                + encbuf[lenhead:len(encbuf)-lentail]))
+            headers = {}
```
ossman: Probably. I seem to recall it was messy though and I didn't feel it was worth the effort. I can have another look.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-
-    def handle(self):
-        # When using run_once, we have a single process, so
-        # we cannot loop in BaseHTTPRequestHandler.handle; we
-        # must return and handle new connections
-        if self.run_once:
-            self.handle_one_request()
-        else:
-            SimpleHTTPRequestHandler.handle(self)
+    def recv(self):
```
ossman: Possibly. For now I figured SSLSocket was more familiar.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-                sock.bind(addrs[0][4])
-                sock.listen(100)
-        else:
-            sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-            sock.connect(unix_socket)
+        # May have been called to flush out a close
+        if self._received_close:
+            self._flush()
+            return ''.encode("ascii")
```
ossman: An empty buffer is usually how `recv()` indicates a nicely closed socket. Anything else seems more like an abnormal condition. `None` can also be the result of a stray `return`.

websockify/websocket.py
```diff
@@ -1,1109 +1,825 @@
...
-    def start_server(self):
-        """
-        Daemonize if requested. Listen for for connections. Run
-        do_handshake() method for each connection. If the connection
-        is a WebSockets client then call new_websocket_client() method (which must
-        be overridden) for each new client connection.
+            self._recv_buffer += data
+
+            # Support for SSLSocket like objects
+            if hasattr(self.socket, "pending"):
```
ossman: That sounds like premature optimisation. :)
This is more readable IMO, and we have no reason to believe `hasattr` is slow.

websockify/websocket.py
ossman: Yeah, it started off as sending then changed. :)

websockify/websocket.py
ossman: > might want to name this payload_length so as to not confuse it with the previous definition of length

Huh? It is not the payload length, but rather the length of the entire packet.

> For that matter, why are we using a dict and not a named tuple or class with __slots__ set here? It's probably more efficient.

dicts are very fast, so I doubt we'll save much. :)

websockify/websocketserver.py
ossman: Great. I'll nuke it. :)

websockify/websocketserver.py
ossman: That would deviate from the SSLSocket API, which is probably not worth it for such a minor thing.

websockify/websocket.py
ossman: I've added `test/echo_client.py` as an example. :)

ossman:
Should be all fixed now.

websockify/websocket.py
DirectXMan12: that's fair, comment closed

websockify/websocket.py
DirectXMan12: > Huh? It is not the payload length, but rather the length of the entire packet.

whoops, I typoed the comment -- what I meant was it used to be just the payload length, so it should probably be `frame_length` or something like that to not confuse things.

> dicts are very fast, so I doubt we'll save much. :)

Sure, but an named tuple is self-documenting in terms of fields and prevents you from accidentally setting different fields.

DirectXMan12:
I'll do one last pass of the recent changes on Monday and then it's probably good to merge.

DirectXMan12:
Fix the conflicts and I'll merge this.

ossman:
Fixed

CendioOssman:
I need to fix some more things, so I'll go ahead and merge this now.

PR #251, opened 2016-10-10
TCP_KEEPCNT doesn't exist in non-linux

thomasgoirand:
The TCP_KEEPCNT option for sockets only work with the Linux kernel,
this isn't available for example in FreeBSD and Hurd, which makes the
package fail to build on these platforms. See Debian bug here:

https://bugs.debian.org/840035


DirectXMan12:
Thanks!

PR #262, opened 2017-01-16
Disable SSL3 in Python wrap_socket

wiebeytec:
Against POODLE security vulnerability.

Note: breaks Python2 support. Websockify runs fine in Python3, which is even available on Ubuntu 12.04, so support for Python2 should probably be dropped.

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -833,11 +833,12 @@ class WebSocketServer(object):
                                   % self.cert)
             retsock = None
             try:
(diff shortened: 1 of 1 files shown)


CendioOssman:
Duplicate/overlap of #190?

DirectXMan12:
> Duplicate/overlap of #190?

looks like it, closing.

DirectXMan12:
(we can always re-open if need-be)

Issue #280, opened 2017-05-17
Recent breaking changes to websockets/ws break websockify.js


samfrances:
See: https://github.com/websockets/ws/pull/1099

The `client.upgradeReq` property has been removed in the `ws` module, leading websockify.js to crash at https://github.com/novnc/websockify/blob/master/other/js/websockify.js#L31

Error message:
```
WebSocket settings:
    - proxying from :6080 to localhost:5900
    - Running in unencrypted HTTP (ws://) mode
/home/sam/websockify_node/websockify.js:31
    console.log(client.upgradeReq.url);
                                 ^

TypeError: Cannot read property 'url' of undefined
    at WebSocketServer.new_client (/home/sam/websockify_node/w
ebsockify.js:31:34)
    at emitTwo (events.js:106:13)
    at WebSocketServer.emit (events.js:191:7)
    at handleUpgrade (/home/sam/websockify_node/node_modules/w
s/lib/WebSocketServer.js:88:16)
    at WebSocketServer.completeUpgrade (/home/sam/websockify_n
ode/node_modules/ws/lib/WebSocketServer.js:270:5)
    at WebSocketServer.handleUpgrade (/home/sam/websockify_nod
e/node_modules/ws/lib/WebSocketServer.js:197:10)
    at Server.WebSocketServer._ultron.on (/home/sam/websockify
_node/node_modules/ws/lib/WebSocketServer.js:87:14)
    at emitThree (events.js:116:13)
    at Server.emit (events.js:194:7)
    at onParserExecuteCommon (_http_server.js:409:14)
```

PR #281, opened 2017-05-17
Fix websockify.js bug caused by upgrade to ws version 3

samfrances:
Fix for https://github.com/novnc/websockify/issues/280

Fixes error in websockify.js caused by upgrade to version 3 of the `ws` module, which contains breaking changes. See also: https://github.com/websockets/ws/pull/1104

### DIFF
--- a/other/js/websockify.js
+++ b/other/js/websockify.js
@@ -28,7 +28,7 @@ var argv = require('optimist').argv,
 // Handle new WebSocket client
 new_client = function(client, req) {
     var clientAddr = client._socket.remoteAddress, log;
-    console.log(req.url);
+    console.log(req ? req.url : client.upgradeReq.url);
     log = function (msg) {
         console.log(' ' + clientAddr + ': '+ msg);
     };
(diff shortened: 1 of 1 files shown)


other/js/websockify.js
```diff
@@ -26,9 +26,9 @@ var argv = require('optimist').argv,
 
 
 // Handle new WebSocket client
-new_client = function(client) {
+new_client = function(client, req) {
     var clientAddr = client._socket.remoteAddress, log;
-    console.log(client.upgradeReq.url);
+    console.log(req.url);
```
lpinca: It is probably better to use something like this:

```js
console.log(req ?  req.url : client.upgradeReq.url);
```

so this will work with all `ws` versions.

samfrances: Thanks. Changed accordingly in https://github.com/novnc/websockify/pull/281/commits/141ba0088ea80cca9d14fa9adaf5f9ca5a8b6d8a

PR #282, opened 2017-05-22
Add authentication plugin capabilities to node version of websockify

samfrances:
I have adapted the node.js version of Websockify, to take advantage of the `ws` library's authorization capabilities. Authorisation plugins can now be specified using the `--auth-plugin` and `--auth-source` command line arguments. `--auth-plugin` specifies the `<module>.<member>` function to use as an auth plugin. `--auth-source` provides additional configuration as a string, which is passed as the first argument to the plugin function. This command line interface mirrors that in the Python version of websockify.

Authorisation plugins are functions that take a `source` argument, and return a function that can be used as the `verifyClient` option to the `ws.WebSocket.Server` class from https://github.com/websockets/ws.

The following abridged extract from the `ws` module documentation at https://github.com/websockets/ws/blob/master/doc/ws.md should clarify:

> 
> ### new WebSocket.Server(options[, callback])
> 
> - `options` {Object}
>   ...
>   - `verifyClient` {Function} A function which can be used to validate incoming
>     connections. See description below.
>
>   ...
> - `callback` {Function}
> 
> Create a new server instance. One of `port`, `server` or `noServer` must be
> provided or an error is thrown.
> 
> 
> If `verifyClient` is not set then the handshake is automatically accepted. If
> it is is provided with a single argument then that is:
> 
> - `info` {Object}
>   - `origin` {String} The value in the Origin header indicated by the client.
>   - `req` {http.IncomingMessage} The client HTTP GET request.
>   - `secure` {Boolean} `true` if `req.connection.authorized` or
>     `req.connection.encrypted` is set.
> 
> The return value (Boolean) of the function determines whether or not to accept
> the handshake.
> 
> if `verifyClient` is provided with two arguments then those are:
> 
> - `info` {Object} Same as above.
> - `cb` {Function} A callback that must be called by the user upon inspection
>   of the `info` fields. Arguments in this callback are:
>   - `result` {Boolean} Whether or not to accept the handshake.
>   - `code` {Number} When `result` is `false` this field determines the HTTP
>     error status code to be sent to the client.
>   - `name` {String} When `result` is `false` this field determines the HTTP
>     reason phrase.

### DIFF
--- a/other/js/auth_plugin_examples.js
+++ b/other/js/auth_plugin_examples.js
@@ -10,9 +10,20 @@
  */
 
 const querystring = require('querystring');
-fs = require('fs');
+const fs = require('fs');
 
 function urlTokenMatch(url, token, verbose=false) {
+    /**
+     * Parse the url path, extract the `token` querystring value, and check if
+     * it matches the token argument. If verbose is set to true, log messages
+     * are enabled.
+     *
+     * Args:
+     *      url (string): the path section of the URL
+     *      token (string): the token which the token provided in the URL should
+     *          match
+     *      verbose (boolean): If True, extra console.log messages will be output
+     */
     let splitUrl = url.split("?")
     if (splitUrl.length !== 2) {
         if (verbose) {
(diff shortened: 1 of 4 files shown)


CendioOssman:
You've added two merges at the end there, which isn't very nice. Other than that it looks good to me.

samfrances:
That was just merging in the latest changes from master. I can file another pull request without those commits if you would prefer.

samfrances:
I have created an identical pull request but without the unwanted merge requests. See: https://github.com/novnc/websockify/pull/285

CendioOssman:
Great. And FYI, you don't need to create a new PR to change commits. Just update your branch and github will adjust.

PR #284, opened 2017-05-23
Feature/249 recording in node js

darkpixel:
This implements the recording feature in the node version of websockify.

I don't believe it is ready to be merged without addressing and/or fixing the following:

* Recording file names are ugly.  I just grabbed the ISO date and used it as a filename.  I'm sure there's a better option.  The python version uses a base filename and adds a incremental number potentially overwriting existing files.  Maybe use a hash of the source IP and date?

* Should I change the parameter from accepting a directory to accepting a 'base' filename?  (If not, I should probably add error handling / create the directory)

* The decodeBuffer function was basically a guess at what characters are valid/invalid.  For example, the current version of the function will encode a period, where the Python version does not.  It's probably not a big deal, but maybe it would be nice to produce output that more closely resembles the Python version.

I appreciate any and all feedback--just wanted to get the conversation rolling.

### DIFF
--- a/other/js/websockify.js
+++ b/other/js/websockify.js
@@ -30,7 +30,7 @@ new_client = function(client, req) {
     var clientAddr = client._socket.remoteAddress, log;
     var start_time = new Date().getTime();
 
-    console.log(client.upgradeReq.url);
+    console.log(req ? req.url : client.upgradeReq.url);
     log = function (msg) {
         console.log(' ' + clientAddr + ': '+ msg);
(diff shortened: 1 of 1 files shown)


CendioOssman:
Looks pretty alright. Please merge these to a single commit and we can probably get it in.

darkpixel:
I'll submit a new PR.

darkpixel:
Created PR #306

PR #285, opened 2017-05-24
Add authentication plugin capabilities to node version of websockify (v2)

samfrances:
Duplicate of pull request https://github.com/novnc/websockify/pull/282 , but without the unwanted merge commits.

------------------------------------------------

I have adapted the node.js version of Websockify, to take advantage of the `ws` library's authorization capabilities. Authorisation plugins can now be specified using the `--auth-plugin` and `--auth-source` command line arguments. `--auth-plugin` specifies the `<module>.<member>` function to use as an auth plugin. `--auth-source` provides additional configuration as a string, which is passed as the first argument to the plugin function. This command line interface mirrors that in the Python version of websockify.

Authorisation plugins are functions that take a `source` argument, and return a function that can be used as the `verifyClient` option to the `ws.WebSocket.Server` class from https://github.com/websockets/ws.

The following abridged extract from the `ws` module documentation at https://github.com/websockets/ws/blob/master/doc/ws.md should clarify:

> 
> ### new WebSocket.Server(options[, callback])
> 
> - `options` {Object}
>   ...
>   - `verifyClient` {Function} A function which can be used to validate incoming
>     connections. See description below.
>
>   ...
> - `callback` {Function}
> 
> Create a new server instance. One of `port`, `server` or `noServer` must be
> provided or an error is thrown.
> 
> 
> If `verifyClient` is not set then the handshake is automatically accepted. If
> it is is provided with a single argument then that is:
> 
> - `info` {Object}
>   - `origin` {String} The value in the Origin header indicated by the client.
>   - `req` {http.IncomingMessage} The client HTTP GET request.
>   - `secure` {Boolean} `true` if `req.connection.authorized` or
>     `req.connection.encrypted` is set.
> 
> The return value (Boolean) of the function determines whether or not to accept
> the handshake.
> 
> if `verifyClient` is provided with two arguments then those are:
> 
> - `info` {Object} Same as above.
> - `cb` {Function} A callback that must be called by the user upon inspection
>   of the `info` fields. Arguments in this callback are:
>   - `result` {Boolean} Whether or not to accept the handshake.
>   - `code` {Number} When `result` is `false` this field determines the HTTP
>     error status code to be sent to the client.
>   - `name` {String} When `result` is `false` this field determines the HTTP
>     reason phrase.

### DIFF
--- a/other/js/auth_plugin_examples.js
+++ b/other/js/auth_plugin_examples.js
@@ -10,9 +10,20 @@
  */
 
 const querystring = require('querystring');
-fs = require('fs');
+const fs = require('fs');
 
 function urlTokenMatch(url, token, verbose=false) {
+    /**
+     * Parse the url path, extract the `token` querystring value, and check if
+     * it matches the token argument. If verbose is set to true, log messages
+     * are enabled.
+     *
+     * Args:
+     *      url (string): the path section of the URL
+     *      token (string): the token which the token provided in the URL should
+     *          match
+     *      verbose (boolean): If True, extra console.log messages will be output
+     */
     let splitUrl = url.split("?")
     if (splitUrl.length !== 2) {
         if (verbose) {
@@ -34,6 +45,11 @@ function urlTokenMatch(url, token, verbose=false) {
 }
 
 exports.tokenAuth = function tokenAuth(source) {
+    /**
+     * Authorisation plugin which validates the token query parameter against
+     * a token provided as the argument to the --auth-source command line
+     * argument
+     */
     return function(info) {
         let token = source;
         return urlTokenMatch(info.req.url, token, true);
@@ -41,6 +57,12 @@ exports.tokenAuth = function tokenAuth(source) {
 }
 
 exports.tokenAuthEnv = function tokenAuthEnv(source) {
+    /**
+     * Authorisation plugin which validates the token query parameter against
+     * a token which is the value of an environment variable. The name of this
+     * environment variable is specified as the argument to the command line
+     * argument --auth-source
+     */
     return function(info) {
         let token = process.env[source];
         return urlTokenMatch(info.req.url, token, true);
@@ -48,6 +70,11 @@ exports.tokenAuthEnv = function tokenAuthEnv(source) {
 }
 
 exports.tokenAuthFile = function tokenEnvFile(source) {
+    /**
+     * Authorisation plugin which validates the token query parameter against a
+     * token which is contained in a text file, the path to which is specified
+     * as the value of the --auth-source command line argument
+     */
     return function(info, cb) {
         fs.readFile(source, 'utf8', function(err, data) {
             if (err) {
--- a/other/js/auth_plugin_examples.js
+++ b/other/js/auth_plugin_examples.js
@@ -47,8 +47,8 @@ function urlTokenMatch(url, token, verbose=false) {
 exports.tokenAuth = function tokenAuth(source) {
     /**
      * Authorisation plugin which validates the token query parameter against
-     * a token provided as the argument to the --auth-source command line
-     * argument
+     * a token provided as the argument to the `--auth-source` command line
+     * argument.
      */
     return function(info) {
         let token = source;
@@ -61,7 +61,7 @@ exports.tokenAuthEnv = function tokenAuthEnv(source) {
      * Authorisation plugin which validates the token query parameter against
      * a token which is the value of an environment variable. The name of this
      * environment variable is specified as the argument to the command line
-     * argument --auth-source
+     * argument `--auth-source`
      */
     return function(info) {
         let token = process.env[source];
@@ -73,7 +73,7 @@ exports.tokenAuthFile = function tokenEnvFile(source) {
     /**
      * Authorisation plugin which validates the token query parameter against a
      * token which is contained in a text file, the path to which is specified
-     * as the value of the --auth-source command line argument
+     * as the value of the `--auth-source` command line argument
      */
     return function(info, cb) {
         fs.readFile(source, 'utf8', function(err, data) {
--- a/other/js/auth_plugin_examples.js
+++ b/other/js/auth_plugin_examples.js
@@ -1,0 +1,27 @@
+/*
+ * An auth plugin must be a function which returns a function conforing to the
+ * requirements of the `verifyClients` option on ws.WebSocket.Server.
+ *
+ * See: https://github.com/websockets/ws/blob/master/doc/ws.md
+ *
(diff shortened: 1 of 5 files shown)


other/js/auth_plugin_examples.js
```diff
@@ -0,0 +1,90 @@
+/*
+ * An auth plugin must be a function which returns a function conforing to the
```
kanaka: conforing -> conforming

other/js/auth_plugin_examples.js
```diff
@@ -0,0 +1,90 @@
...
+
+const querystring = require('querystring');
+const fs = require('fs');
+
+function urlTokenMatch(url, token, verbose=false) {
+    /**
+     * Parse the url path, extract the `token` querystring value, and check if
+     * it matches the token argument. If verbose is set to true, log messages
+     * are enabled.
+     *
```
kanaka: The "token" query parameter already has a different meaning in the main python implementation which is used for selecting among multiple target hosts: https://github.com/novnc/websockify/wiki/Token-based-target-selection. The python implementation also supports basic authentication via the Authorization header. I think it would be better to have an example of basic auth as that would probably be more generally applicable. But certainly it shouldn't be using "token" for authentication.

other/js/websockify.js
```diff
@@ -161,7 +162,23 @@ if (argv.cert) {
     console.log("    - Running in unencrypted HTTP (ws://) mode");
     webServer = http.createServer(http_request);
 }
+
+if (argv["auth-plugin"]) {
+    let auth_plugin_arg = argv["auth-plugin"].split(".")
+    let plugin_name = auth_plugin_arg.pop();
+    let module_path = auth_plugin_arg.join(".");
+
+    let auth_plugin = require(module_path)[plugin_name];
```
kanaka: In node it's not unusual for the module itself to have a single export (exports = myFunc). That case should probably be supported.

other/js/websockify.js
```diff
@@ -161,7 +162,23 @@ if (argv.cert) {
...
+    let auth_source = argv["auth-source"] || undefined;
+    websocket_server_opts = {
+        server: webServer,
+        verifyClient: auth_plugin(auth_source)
```
kanaka: What about the case (that is symmetrical to the python version) where auth_plugin is a class function and not just a plain function returning the auth function?

I think it would actually be more idiomatic JavaScript to have the auth_source bound as the first argument of an authorizer function rather than generating a new function in the auth_plugin itself. For example:

```
verifyClient: auth_plugin.bind(auth_plugin, auth_source)
```

Obviously, the structure of your example authentication function would change (one less layer of abstraction).

samfrances:
Thanks for the review points, @kanaka . I'll get onto those ASAP.

other/js/auth_plugin_examples.js
samfrances: Fixed in https://github.com/novnc/websockify/pull/285/commits/dc31fab663e7883e5a72ad15f044ffbf4a4e4b18

CendioOssman:
How are the fixes going?

samfrances:
Sorry @CendioOssman , this has been on the back burner for me. I'll get on it this weekend.

other/js/websockify.js
samfrances: Addressed in 5c3ac52. The auth plugin can now either be a class with a `authenticate` method (similar to the Python version), or a factory function which returns an object with an `authenticate` method.

other/js/websockify.js
samfrances: Addressed in https://github.com/novnc/websockify/pull/285/commits/ff1897ee94b2b15bc5cff357e22d36e99f9f1164

other/js/auth_plugin_examples.js
samfrances: Replaced token based auth examples with origin-based examples (these being easier to test out using NoVNC than basic auth): https://github.com/novnc/websockify/pull/285/commits/a543c359438b2368010cff352d07148e1136295e

samfrances:
@CendioOssman @kanaka I believe I have addressed all of the review comments now. See what you think.

CendioOssman:
@kanaka?

kanaka:
@samfrances I would love to be able to review this in detail, but I just haven't been able to find the time lately. @CendioOssman if you have time to do brief review and maybe test it quickly, I have no objection to merging it.

samfrances:
@kanaka Certainly happy to follow up on any issues.

CendioOssman:
I can try to review it, but right now I'm a bit swamped with other things.

other/js/websockify.js
```diff
@@ -92,7 +93,7 @@ http_request = function (request, response) {
 
     var uri = url.parse(request.url).pathname
         , filename = path.join(argv.web, uri);
-    
+
```
samhed: This white space change is not really close to other modified code. I would prefer if you could move it to a separate commit

other/js/auth_plugin_examples.js
samhed: could you squash dc31fab into 7d6f638 ?

other/js/auth_plugin_examples.js
```diff
@@ -47,8 +47,8 @@ function urlTokenMatch(url, token, verbose=false) {
 exports.tokenAuth = function tokenAuth(source) {
     /**
      * Authorisation plugin which validates the token query parameter against
-     * a token provided as the argument to the --auth-source command line
-     * argument
+     * a token provided as the argument to the `--auth-source` command line
+     * argument.
```
samhed: Could you squash the changes from this commit (22ffd2c) into the commit that originally introduced these comments please

other/js/password.txt
```diff
@@ -1 +0,0 @@
-passpass12
```
samhed: Squash this commit (e5e0e5d5686384820a488e0f5fa8e40d942b6c28) with 0e1824c0ff64e008e0cd24de5f1a02f1de6e5394 please

other/js/websockify.js
```diff
@@ -29,9 +29,9 @@ var argv = require('optimist').argv,
 
 
 // Handle new WebSocket client
-new_client = function(client, upgradeReq) {
+new_client = function(client, req) {
     var clientAddr = client._socket.remoteAddress, log;
-    console.log(upgradeReq ? upgradeReq.url : client.upgradeReq.url);
+    console.log(req ? req.url : client.req.url);
```
samhed: You should rebase your changes on master to avoid commits like this one

samfrances:
Closing, as this was from a while ago and I don't think I'll have time to do the requested changes.

PR #293, opened 2017-07-31
Add support for inetd.

michalsrb:
With the --inetd parameter, websockify doesn't require the source_addr and
source_port paramters and it expects that stdin is already opened and listening
socket.

This way websockify can be used with (x)inetd or as a systemd socket-activated
service.


michalsrb:
Apparently some tests failed, but it seems they were failing the same way in previous build already:
https://travis-ci.org/novnc/websockify/builds/243259716

websockify/websocketproxy.py
```diff
@@ -457,44 +464,53 @@ def websockify_init():
...
+            opts.listen_host = opts.listen_host.strip('[]')
+        else:
+            opts.listen_host, opts.listen_port = '', arg
 
-    try:    opts.listen_port = int(opts.listen_port)
-    except: parser.error("Error parsing listen port")
+        try:    opts.listen_port = int(opts.listen_port)
+        except: parser.error("Error parsing listen port")
+
+    del opts.inetd
```
CendioOssman: Why?

websockify/websocketproxy.py
```diff
@@ -297,12 +297,17 @@ def started(self):
         else:
             dst_string = "%s:%s" % (self.target_host, self.target_port)
 
+        if self.listen_fd != None:
+            src_string = "fd %d" % self.listen_fd
```
CendioOssman: This isn't terribly helpful for anyone. Perhaps just omit the entire "from ..." part if we don't know it?

websockify/websockifyserver.py
```diff
@@ -371,8 +372,11 @@ def __init__(self, RequestHandlerClass, listen_host='',
 
         # Show configuration
         self.msg("WebSocket server settings:")
-        self.msg("  - Listen on %s:%s",
-                self.listen_host, self.listen_port)
+        if self.listen_fd != None:
+            self.msg("  - Listen on fd %d", self.listen_fd)
```
CendioOssman: Same thing here. Perhaps "Listen for inetd connections" or something like that?

websockify/websocketproxy.py
michalsrb: Because the `opts` are eventually used as parameters of `WebSockifyServer` constructor and it would fail because it doesn't know the `inetd` parameter. This deletion follows the same pattern as for example `opts.target_cfg` few lines above.

websockify/websocketproxy.py
michalsrb: Ok, sure.

websockify/websockifyserver.py
michalsrb: Ok

websockify/websocketproxy.py
CendioOssman: Alright, fair enough.

michalsrb:
Alright, I modified the log messages in the "Add support for inetd." commit.

Issue #207, opened 2015-10-28
Make using letsencrypt.org really easy


kanaka:
To encourage people to use wss/tls encryption, we should make it really simple for people to implement a signed cert using letsencrypt.org (which should be going into general availability in the next couple of months). Perhaps direct automatic integration (since that's one of the goals of Let's Encrypt), but at least, we should document a straightforward process for using it.

DirectXMan12:
:+1: to that

dzil123:
Well, I think that the first step is adding instructions, or even a link, on hows to use letsencrypt, on the [Encrypted Connections wiki](https://github.com/kanaka/websockify/wiki/Encrypted-Connections) page. It would be much simpler than having to manually acceptance certificates in the browser, which is the current recommendation.

samhed:
@dzil123 we would gladly review/accept a PR ;)

hoehermann:
Today, I used letsencrypt's certbot to obtain certificates:

    sudo certbot certonly --standalone -d host.example.com

I am now using websockify with

    --cert=/etc/letsencrypt/live/host.example.com/fullchain.pem --key=/etc/letsencrypt/live/host.example.com/privkey.pem

It won't get any easier than that, I guess.

stevegt:
Right now running novnc/websockify over SSL with unattended deployment requires extra startup scripting, because certbot needs to be run on the host or container after it's up and reachable via DNS, but before novnc or websockify starts.

This would be made easier by adding email and fqdn flags to noVNC/utils/launch.sh.  This would enable those bits to be passed to websockify, where websockify can then call this:

`certbot -n certonly --standalone --agree-tos --email $EMAIL --domains $FQDN`

PR #295, opened 2017-08-29
SSL client validation (certificate-based authentication)

hoehermann:
I don't like passwords. But I love public key authentication. That is why I added SSL client validation to websockify (specifically for use with noVNC). I know this won't make it into upstream due to compatibility issues as discussed in #190, but other people might want to know this feature now exists in at least one fork of Python websockify. For maximum ease of use, I recommend certificates by letsencrypt as mentioned in #207.

The test `test_do_handshake_ssl_error_eof_raises_close_error` fails since the ssl module does not actually accept the fake socket of the FakeSocket type (there is an explicit type check).

### DIFF
--- a/tests/test_websockifyserver.py
+++ b/tests/test_websockifyserver.py
@@ -257,10 +257,36 @@ class WebSockifyServerTestCase(unittest.TestCase):
             sock, '127.0.0.1')
 
     def test_do_handshake_ssl_error_eof_raises_close_error(self):
-        # TODO: re-implement this test.
-        # Test was incompatible with new style socket wrapping offered by
-        # ssl.create_default_context.
-        pass
+        server = self._get_server(daemon=True, ssl_only=0, idle_timeout=1)
+
+        sock = FakeSocket("\x16some ssl data")
+
+        def fake_select(rlist, wlist, xlist, timeout=None):
+            return ([sock], [], [])
+
+        def fake_wrap_socket(*args, **kwargs):
+            raise ssl.SSLError(ssl.SSL_ERROR_EOF)
+
+        class fake_create_default_context():
+            def __init__(self, purpose):
+                self.verify_mode = None
+            def load_cert_chain(self, certfile, keyfile):
+                pass
+            def set_default_verify_paths(self):
+                pass
+            def load_verify_locations(self, cafile):
+                pass
+            def wrap_socket(self, *args, **kwargs):
+                raise ssl.SSLError(ssl.SSL_ERROR_EOF)
+
+        self.stubs.Set(select, 'select', fake_select)
+        # for recent versions of python
+        self.stubs.Set(ssl, 'create_default_context', fake_create_default_context)
+        # for fallback for old versions of python
+        self.stubs.Set(ssl, 'wrap_socket', fake_wrap_socket)
+        self.assertRaises(
+            websockifyserver.WebSockifyServer.EClose, server.do_handshake,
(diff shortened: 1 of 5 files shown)


samhed:
The python requirement comment doesn't mean it's impossible. Just make sure that it doesn't break on systems with old python. You can add an 'if' that checks the version and just return if its too old.

hoehermann:
Fallback is added in 8cb3acd.

websockify/websocketproxy.py
```diff
@@ -392,6 +406,12 @@ def websockify_init():
             help="disallow non-encrypted client connections")
     parser.add_option("--ssl-target", action="store_true",
             help="connect to SSL target as SSL client")
+    parser.add_option("--verify-client", action="store_true",
+            help="require encrypted client to present a valid certificate")
+    parser.add_option("--cafile", metavar="FILE",
+            help="file of concatenated certificates of authorities trusted "
+            "for validating clients (only effective with --verify-client). "
+            "If omitted, system default list of CAs is used.")
```
CendioOssman: This isn't completely true, is it? The code suggests that system CAs will always be used.

websockify/auth_plugins.py
```diff
@@ -82,7 +82,7 @@ def authenticate(self, headers, target_host, target_port):
         if origin is None or origin not in self.source:
             raise InvalidOriginError(expected=self.source, actual=origin)
             
-class ClientCertAuth(object):
+class ClientCertCNAuth(object):
```
CendioOssman: Please merge this commit with the earlier ones.

websockify/websockifyserver.py
```diff
@@ -541,16 +541,31 @@ def do_handshake(self, sock, address):
...
+                    if self.verify_client:
+                        context.verify_mode = ssl.CERT_REQUIRED
+                        context.set_default_verify_paths()
+                        if self.cafile:
+                            context.load_verify_locations(cafile=self.cafile)
+                    retsock = context.wrap_socket(
+                            sock,
+                            server_side=True)
+                except AttributeError as ae:
+                    if str(ae) != "'module' object has no attribute 'create_default_context'":
```
CendioOssman: This can be done a lot cleaner. E.g.
```python
if hasattr(ssl, 'create_default_context'):
```

CendioOssman: When fixed this should also be merged in to the original commit.

tests/test_websockifyserver.py
```diff
@@ -257,10 +257,36 @@ def fake_select(rlist, wlist, xlist, timeout=None):
...
+                raise ssl.SSLError(ssl.SSL_ERROR_EOF)
+
+        self.stubs.Set(select, 'select', fake_select)
+        # for recent versions of python
+        self.stubs.Set(ssl, 'create_default_context', fake_create_default_context)
+        # for fallback for old versions of python
+        self.stubs.Set(ssl, 'wrap_socket', fake_wrap_socket)
+        self.assertRaises(
+            websockifyserver.WebSockifyServer.EClose, server.do_handshake,
+            sock, '127.0.0.1')
```
CendioOssman: Merge this in to the first commit as well.

websockify/websocketproxy.py
hoehermann: Indeed. I should move the call to `context.set_default_verify_paths()` into the `else` alternative.

hoehermann:
Requested changes made in new pull request #308.

PR #299, opened 2017-09-26
On demand systemd or upstart socket activation

valia0906:
On demand systemd or upstart socket activation.


websockify/websockifyserver.py
```diff
@@ -425,8 +434,13 @@ def socket(host, port=None, connect=False, prefer_ipv6=False,
             raise Exception("SSL only supported in connect mode (for now)")
         if not connect:
             flags = flags | socket.AI_PASSIVE
-
-        if not unix_socket:
+        
+        # Starting systemd or upstart socket
+        if host == "socketfromfd":
```
CendioOssman: If you create a new variable then you should be able to use it here. I.e. `if this.listen_fd`.

PR #303, opened 2017-10-12
Send Content-Type headers with static files

kosmasgiannis:
Fixes a problem that occurs in Chrome 61 where the following error message appears in the console:

'Failed to load module script: The server responded with a non-JavaScript MIME type of "".
Strict MIME type checking is enforced for module scripts per HTML spec.'


CendioOssman:
Looks good. Thanks.

PR #313, opened 2017-11-10
Do not confuse an empty message with a closed connection

### LINKED ISSUE #312: WebSocket message with empty payload misinterpreted as close message
As of commit 8a697622495fd319582cd1c604e7eb2cc0ac0ef6, a WebSocket message with payload length zero causes the connection to be closed (“Client closed connection”). This seems to be inconsistent with the specification and with the previous behavior of websockify.

andersk:
Fixes #312.

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -343,9 +343,9 @@ class WebSocket(object):
         """Read data from the WebSocket.
 
         This will return any available data on the socket. If the
-        socket is closed then an empty buffer will be returned. The
-        reason for the close is found in the 'close_code' and
-        'close_reason' properties.
+        socket is closed then None will be returned. The reason for
+        the close is found in the 'close_code' and 'close_reason'
+        properties.
 
         Unlike recvmsg() this method may return data from more than one
         WebSocket message. It is however not guaranteed to return all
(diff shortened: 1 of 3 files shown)


CendioOssman:
Argh. I was trying to make it match the behaviour of `Socket` as much as possible. But you are right, WebSockets are message oriented.

However, we have both `recv()` and `recvmsg()`. Perhaps only `recvmsg()` should be changed to return `None`? That would keep the API more familiar for those using things as a normal socket, and still allow full functionality for those who try to do message based communication?

andersk:
We could filter out empty messages from `recv`, so that it can use the empty string unambiguously for the closed case:

```python
    def recv(self):
        msg = self.recvmsg()
        while msg == '':
            msg = self.recvmsg()
        return '' if msg is None else msg
```

But then we also need two versions of `pending`, one that checks whether messages are available for use with `recvmsg`, and one that checks whether _nonempty_ messages are available for use with `recv`.

CendioOssman:
You are right, the rabbit hole just goes on. And since we have `pending()` we're already different from normal `Socket`s anyway. So let's go with your initial plan of `None` everywhere.

However, to avoid surprising people, could you update the docstring for `recv()` to mention that the message oriented nature of WebSockets could result in getting an empty buffer?

andersk:
Sure, how’s this?

CendioOssman:
Great, thanks. :)

PR #329, opened 2018-02-07
C websockify: use openssl base64 encode/decode functions

### LINKED ISSUE #165: b64_ntop() / b64_pton() not universally available on all C libraries
I am trying to cross-compile the C version of websockify for an embedded system that uses uclibc instead of glibc.
However it seems b64_ntop() / b64_pton() is not available with that.

```
websocket.o: In function `encode_hixie':
websocket.c:(.text+0x3f0): undefined reference to `b64_ntop'
websocket.o: In function `decode_hixie':
websocket.c:(.text+0x4e8): undefined reference to `b64_pton'
websocket.o: In function `encode_hybi':
websocket.c:(.text+0x614): undefined reference to `b64_ntop'
websocket.o: In function `decode_hybi':
websocket.c:(.text+0x86c): undefined reference to `b64_pton'
websocket.o: In function `do_handshake':
websocket.c:(.text+0x1224): undefined reference to `b64_ntop'
collect2: error: ld returned 1 exit status
```

Please consider bundling those methods locally with websockify.

pesintta:
b64_pton and b64_ntop functions are not portable and cannot be found in
all C library implementations (e.g. uClibc, musl).

Since c-websockify already has explicit dependency to openssl it can be
used to replace b64_pton/ntop with versions that are portable without
introducing too much additional code or dependencies.

Fixes #165

### DIFF
--- a/other/Makefile
+++ b/other/Makefile
@@ -4,7 +4,7 @@ CFLAGS += -fPIC
 all: $(TARGETS)
 
 websockify: websockify.o websocket.o
-	$(CC) $(LDFLAGS) $^ -lssl -lcrypto -lresolv -o $@
+	$(CC) $(LDFLAGS) $^ -lssl -lcrypto -o $@
 
 websocket.o: websocket.c websocket.h
 websockify.o: websockify.c websocket.h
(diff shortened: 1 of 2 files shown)


CendioOssman:
Seems reasonable.

PR #344, opened 2018-04-15
 Append mime-type while proxy http service using python2

colin-zhou:
https://github.com/novnc/noVNC/issues/1051

With this piece of code, we can make it work under python2 environment

### DIFF
--- a/websockify/websockifyserver.py
+++ b/websockify/websockifyserver.py
@@ -12,7 +12,7 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import os, sys, time, errno, signal, socket, select, logging
+import os, sys, time, errno, signal, socket, select, logging, mimetypes
 import multiprocessing
 
 # Imports that vary by python version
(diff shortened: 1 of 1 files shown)


samhed:
You can force-push to the same branch to overwrite your old commits, that way you won't have to open new PR's.

colin-zhou:
Thanks, I didn't know this until now.

CendioOssman:
We don't really want to start maintaining a list like this, and this should not be necessary on a properly working system. Under what scenario are you seeing problems?

colin-zhou:
I tried to install the mailcap on Ret Hat, but it still not work. From my perspective, the mainstream http server like nginx (its default mime config), apache(its mod_mime config) and uwsgi(read from /etc/mime.type by default) all provide the mime map by itself to make it robust under different environments, so considering map the mime type by the httpserver itself is not bad.  By the way, I will find the rule cause of noVNC not work properly on Red Hat.

CendioOssman:
Works fine with my RHEL 7 machine here:

```
$ python -c 'import mimetypes; mimetypes.init(); print mimetypes.types_map[".svg"];'
  File "<string>", line 1, in <module>
KeyError: '.svg'
$ sudo yum install mailcap
...
$ python -c 'import mimetypes; mimetypes.init(); print mimetypes.types_map[".svg"];'
image/svg+xml
```

colin-zhou:
You are right, a new centos container with mailcap installed works well. Thanks very much.

PR #346, opened 2018-04-24
Use path in URL to pass VNC server address:port for connecting different VNC servers

QunL:
Use path in URL to pass VNC server address:port for connecting different VNC servers;
The path argument is hashed and encoded for anti-tampering;
User can define group policy of user id only accessing VNC server in the same group;

Our requirement and design can refer https://docs.google.com/document/d/e/2PACX-1vQ-LGFjY3iOXKS926WNlyNBdFcU27mrfhDHXai2EXktkaW-QS1aBAtxELuYt38f1BZ1T4bl3mFh62Op/pub

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,9 @@ and websockify as a browser telnet client (`wstelnet.html`).
 
 These are not necessary for the basic operation.
 
+* Muliti-vncserver: Use path in URL to pass VNC server address:port for
+  connecting different VNC servers.
+
 * Daemonizing: When the `-D` option is specified, websockify runs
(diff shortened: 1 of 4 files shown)


CendioOssman:
This seems like a lot of complexity for something that could already be solved with the existing token plugin. Have you looked at that and checked if it can fulfil your needs?

websockify is a simple server for testing and prototyping. It has no ambitions of being production ready, so in general it is not really the right place for advanced functionality.

CendioOssman:
No response. Closing.

Issue #334, opened 2018-03-09
Logging original client behind reverse proxy


praimmugen:
Hi,
I'm running websockify behind Apache configured as reverse proxy. websockify listens on the loopback interface and connects to the VNC servers using a custom token plugin.

Everything works nicely but I'd like to get info regarding the original client in the logs of websockify for each relevant entry.

I made some changes to the request handler and to `WebSockifyServer` to accomodate this. I'm sharing them in case anybody is interested (using websockify behind a reverse proxy seems to be quite common), and I would greatly appreciate some feedback.

The change to the request handler is quite straightforward (just get the info from the `X-Forwarded-For` header, if present). 

I also wanted to show this in the logs produced by `WebSockifyServer`, for example when no token is found. When this happens an exception gets raised from the request handler, so I added the information to the exception for `WebSockifyServer` to read. It works but I don't particularly like it, can anybody think of a better way to do it?

Thanks!

The log messages are like the following, when communication is proxied:
```
192.168.0.109 (forwarded by 127.0.0.1) - - [08/Mar/2018 17:38:59] "GET /?token=tokenABC HTTP/1.1" 101 -
```

And here's the diff:
```diff
diff --git a/websockify/websocketserver.py b/websockify/websocketserver.py
index 6d4824a..ef525d9 100644
--- a/websockify/websocketserver.py
+++ b/websockify/websocketserver.py
@@ -35,6 +35,13 @@ class WebSocketRequestHandler(BaseHTTPRequestHandler):
     def __init__(self, request, client_address, server):
         BaseHTTPRequestHandler.__init__(self, request, client_address, server)
 
+    def original_client(self):
+        # Get the originating client if running behind reverse proxies.
+        # Defaults to BaseHTTPRequestHandler.address_string() if request
+        # is not proxied
+        fwd_list = self.headers.get("X-Forwarded-For", None)
+        return fwd_list.split(",")[0] if fwd_list else self.address_string()
+
     def handle_one_request(self):
         """Extended request handler
 
diff --git a/websockify/websockifyserver.py b/websockify/websockifyserver.py
index f6c067c..366556c 100644
--- a/websockify/websockifyserver.py
+++ b/websockify/websockifyserver.py
@@ -99,13 +99,19 @@ class WebSockifyRequestHandler(WebSocketRequestHandler, SimpleHTTPRequestHandler
 
         WebSocketRequestHandler.__init__(self, req, addr, server)
 
-    def log_message(self, format, *args):
[...]

PR #351, opened 2018-05-03
[Python] Cmd line parameter to log proxied clients

praimmugen:
When running behind a reverse proxy, websockify doesn't currently read and log anything that could provide information on the originating client.

This code adds a startup flag (`--log-proxied-client`) to enable logging of originating client plus intermediate proxies. Information is retrieved from the "X-Forwarded-For" header (if present). The immediate client would be logged too at the end of the content of the header, separated by a comma.

Eg:
* default message: `127.0.0.1: Token 'abcd' not found`
* using the `--log-proxied-client` option: `192.168.0.51,127.0.0.1: Token 'abcd' not found`

See also https://github.com/novnc/websockify/issues/334

### DIFF
--- a/websockify/websocketserver.py
+++ b/websockify/websocketserver.py
@@ -17,6 +17,7 @@ except ImportError:
 
 from websockify.websocket import WebSocket, WebSocketWantReadError, WebSocketWantWriteError
 
+
 class WebSocketRequestHandler(BaseHTTPRequestHandler):
     """WebSocket request handler base class.
 
@@ -34,6 +35,13 @@ class WebSocketRequestHandler(BaseHTTPRequestHandler):
 
     def __init__(self, request, client_address, server):
         BaseHTTPRequestHandler.__init__(self, request, client_address, server)
+
+    def address_string(self, show_proxied=False):
+        immediate_client = super(WebSocketRequestHandler, self).address_string()
+        fwd_for = self.headers.get("X-Forwarded-For", None)
+        if show_proxied and fwd_for:
+            return "{},{}".format(fwd_for, immediate_client)
+        return immediate_client
 
     def handle_one_request(self):
         """Extended request handler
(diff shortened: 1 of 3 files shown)


websockify/websocketserver.py
```diff
@@ -17,6 +17,7 @@
 
 from websockify.websocket import WebSocket, WebSocketWantReadError, WebSocketWantWriteError
 
+
```
CendioOssman: Noise. :)

websockify/websocketserver.py
```diff
@@ -35,6 +36,13 @@ class WebSocketRequestHandler(BaseHTTPRequestHandler):
     def __init__(self, request, client_address, server):
         BaseHTTPRequestHandler.__init__(self, request, client_address, server)
 
+    def address_string(self, show_proxied=False):
+        immediate_client = super(WebSocketRequestHandler, self).address_string()
+        fwd_for = self.headers.get("X-Forwarded-For", None)
+        if show_proxied and fwd_for:
+            return "{},{}".format(fwd_for, immediate_client)
```
CendioOssman: This format string isn't compatible with Python 2.6, which is our current baseline.

websockify/websockifyserver.py
```diff
@@ -88,24 +88,29 @@ def __init__(self, req, addr, server):
         self.daemon = getattr(server, "daemon", False)
         self.record = getattr(server, "record", False)
         self.run_once = getattr(server, "run_once", False)
-        self.rec        = None
+        self.rec = None
```
CendioOssman: More noise

websockify/websockifyserver.py
```diff
@@ -671,11 +687,17 @@ def top_new_client(self, startsock, address):
         try:
             try:
                 client = self.do_handshake(startsock, address)
-            except self.EClose:
+            except self.EClose as e:
                 _, exc, _ = sys.exc_info()
                 # Connection was not a WebSockets connection
                 if exc.args[0]:
-                    self.msg("%s: %s" % (address[0], exc.args[0]))
+                    orig_client = getattr(e, 'orig_client', None)
```
CendioOssman: I don't think we should start piggy-backing extra information on the exception object. Better to add this information to the connection object somewhere.

CendioOssman:
Ping @praimmugen.

praimmugen:
Sorry, had to divert my attention to something else. Will get back to this on the weekend - I see there are some conflicts, too, so better not to wait too much.

samhed:
No response.

PR #353, opened 2018-05-21
Auth changes

josedpedroso:
Added --web-auth option to require authentication to access the webserver.
BasicHTTPAuth plugin now issues 401 on bad credentials to allow the user to try again.

### DIFF
--- a/websockify/auth_plugins.py
+++ b/websockify/auth_plugins.py
@@ -40,35 +40,41 @@ class BasicHTTPAuth(object):
         auth_header = headers.get('Authorization')
         if auth_header:
             if not auth_header.startswith('Basic '):
-                raise AuthenticationError(response_code=403)
+                self.auth_error()
 
             try:
                 user_pass_raw = base64.b64decode(auth_header[6:])
             except TypeError:
-                raise AuthenticationError(response_code=403)
+                self.auth_error()
 
             try:
                 # http://stackoverflow.com/questions/7242316/what-encoding-should-i-use-for-http-basic-authentication
                 user_pass_as_text = user_pass_raw.decode('ISO-8859-1')
             except UnicodeDecodeError:
-                raise AuthenticationError(response_code=403)
+                self.auth_error()
 
             user_pass = user_pass_as_text.split(':', 1)
             if len(user_pass) != 2:
-                raise AuthenticationError(response_code=403)
+                self.auth_error()
 
             if not self.validate_creds(*user_pass):
-                raise AuthenticationError(response_code=403)
+                self.demand_auth()
 
         else:
-            raise AuthenticationError(response_code=401,
-                                      response_headers={'WWW-Authenticate': 'Basic realm="Websockify"'})
+            self.demand_auth()
 
     def validate_creds(self, username, password):
         if '%s:%s' % (username, password) == self.src:
             return True
         else:
             return False
+
+    def auth_error(self):
+        raise AuthenticationError(response_code=403)
+
+    def demand_auth(self):
+        raise AuthenticationError(response_code=401,
+                                  response_headers={'WWW-Authenticate': 'Basic realm="Websockify"'})
 
 class ExpectOrigin(object):
     def __init__(self, src=None):
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -373,19 +373,18 @@ def _subprocess_setup():
     signal.signal(signal.SIGPIPE, signal.SIG_DFL)
 
 
-def logger_init():
+def websockify_init():
+    # Setup basic logging to stderr.
     logger = logging.getLogger(WebSocketProxy.log_prefix)
     logger.propagate = False
     logger.setLevel(logging.INFO)
-    h = logging.StreamHandler()
-    h.setLevel(logging.DEBUG)
-    h.setFormatter(logging.Formatter("%(message)s"))
-    logger.addHandler(h)
+    stderr_handler = logging.StreamHandler()
+    stderr_handler.setLevel(logging.DEBUG)
+    log_formatter = logging.Formatter("%(message)s")
+    stderr_handler.setFormatter(log_formatter)
+    logger.addHandler(stderr_handler)
 
-
-def websockify_init():
-    logger_init()
-
+    # Setup optparse.
     usage = "\n    %prog [options]"
     usage += " [source_addr:]source_port [target_addr:target_port]"
     usage += "\n    %prog [options]"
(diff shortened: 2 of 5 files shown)


josedpedroso:
Not sure if syslog support is desirable; ignore `da185b8` if so.

websockify/websocketproxy.py
```diff
@@ -56,7 +56,8 @@ def send_auth_error(self, ex):
         self.end_headers()
     
     def validate_connection(self):
-        if self.server.token_plugin:
+        # don't use token for non-ws requests
+        if self.server.token_plugin and self.headers.get('upgrade') and self.headers.get('upgrade').lower() == 'websocket':
```
CendioOssman: This is just messy. I think it's better to split things up instead. Create a new `auth_connection()` or something like that.

websockify/websocketproxy.py
```diff
@@ -90,6 +90,20 @@ def validate_connection(self, no_tokens=False):
                 self.send_auth_error(ex)
                 raise
 
+    def do_GET(self):
```
CendioOssman: Why the move? I made more sense to have it in `WebsockifyServer` to keep symmetry with the existing call to `validate_connection()`.

websockify/auth_plugins.py
```diff
@@ -40,36 +40,42 @@ def authenticate(self, headers, target_host, target_port):
...
             return True
         else:
             return False
 
+    def auth_error(self):
+        raise AuthenticationError(response_code=403)
+
+    def demand_auth(self):
+        raise AuthenticationError(response_code=401,
+                                  response_headers={'WWW-Authenticate': 'Basic realm="Websockify"'})
```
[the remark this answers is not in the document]
josedpedroso: I didn't know exactly what the difference between 401 and 403 was, before looking it up. Having these 2 methods makes the distinction clear.

websockify/websocketproxy.py
josedpedroso: The problem is that I was calling validate_connection(no_tokens=True) in WebSockifyRequestHandler, but the behavior for no_tokens=True was implemented in ProxyRequestHandler, a subclass of WebSockifyRequestHandler. It seemed wrong to make WebSockifyRequestHandler expect any such thing of a subclass, so I just moved everything to ProxyRequestHandler, with the bonus that now the actual functionality for --web-auth is all there in one place.

websockify/websocketproxy.py
```diff
@@ -457,38 +456,74 @@ def websockify_init():
             "directory containing configuration files of this form "
             "(DEPRECATED: use `--token-plugin TokenFile --token-source "
             " path/to/token/file` instead)")
-    parser.add_option("--token-plugin", default=None, metavar="PLUGIN",
-                      help="use the given Python class to process tokens "
-                           "into host:port pairs")
+    parser.add_option("--token-plugin", default=None, metavar="CLASS",
+                      help="use a Python class, usually one from token_plugins.py, "
+                           "such as TokenFile, to process tokens into host:port pairs")
```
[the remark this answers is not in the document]
josedpedroso: Agreed, I'll try to take care of that later today.

websockify/websocketproxy.py
josedpedroso: Agreed it was messy, but this is not how it ended up after the latter commits.

websockify/websocketproxy.py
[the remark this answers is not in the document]
josedpedroso: Done.

websockify/websocketproxy.py
CendioOssman: You still had to add the exception via the extra parameter. So I still say split things up for clarity.

websockify/websocketproxy.py
CendioOssman: That problem will be avoided if we don't need the extra parameter.

websockify/websocketproxy.py
```diff
@@ -456,20 +456,20 @@ def websockify_init():
             "directory containing configuration files of this form "
             "(DEPRECATED: use `--token-plugin TokenFile --token-source "
             " path/to/token/file` instead)")
-    parser.add_option("--token-plugin", default=None, metavar="PLUGIN",
-                      help="use the given Python class to process tokens "
-                           "into host:port pairs")
+    parser.add_option("--token-plugin", default=None, metavar="CLASS",
+                      help="use a Python class, usually one from token_plugins.py, "
```
CendioOssman: The syntax used is the Python import syntax, so we should probably use that here as well. I.e. `websockify.token_plugins` rather than `token_plugins.py`.

websockify/websocketproxy.py
josedpedroso: Alright, I'll take care of that later today.

websockify/websocketproxy.py
josedpedroso: Referring to the files by name makes it a bit easier for users to figure out which options are available for --token-plugin and --auth-plugin, especially if they don't know python.

CendioOssman: Possibly. I guess it highlights a need for more detailed documentation than just --help output..

josedpedroso: Yeah, I might write a wiki page about plugins.

CendioOssman: We have this:

https://github.com/novnc/websockify/wiki/Token-based-target-selection

But it might be better to have documentation in the source tree. People might not have access to the wiki when they are configuring things. (or know about it)

josedpedroso: A brief paragraph in the readme explaining where to find the plugins, then?

websockify/websocketproxy.py
josedpedroso: Done.

websockify/websocketproxy.py
josedpedroso: Done.

websockify/websocketproxy.py
josedpedroso: Check the latest commit, I think it's fine now.

CendioOssman:
Great. Please merge the commits so we don't have a bunch of fixup commits and force push things. Then I can have a final look and we should be ready to merge.

josedpedroso:
I think I messed up the merge. I'll have to sort this mess later today. Sorry.

CendioOssman:
`git rebase -i` is your friend. :)

Take your time. Better to get it right.

josedpedroso:
I ended up doing a soft reset to the commit just before my first one and then staging and committing changes properly. I wasn't really getting the results I wanted with the interactive rebase, even reordering and squashing commits, but that could just be my lack of experience with git. Still, I learned a lot more about git than I was expecting today and it seems to have worked, so it's all good, I think.

CendioOssman:
Nice. Looks all good now. Thank you for your work. :)

PR #355, opened 2018-07-02
Enable setting SSL ciphers and SSL options

tomob:
The change adds two options to WebSockifyServer. The first is a list of
SSL ciphers. The second is SSL options (intended use is to force a
specific TLS version).

Those two options allow for greater security of WebSocket Proxy.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -11,7 +11,7 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import signal, socket, optparse, time, os, sys, subprocess, logging, errno
+import signal, socket, optparse, time, os, sys, subprocess, logging, errno, ssl
 try:
     from socketserver import ForkingMixIn
 except ImportError:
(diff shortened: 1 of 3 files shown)


CendioOssman:
Nice. Thank you.

However websockify is primarily used from the command line. So I'd really like to see this hooked up to command line arguments in `WebsocketProxy`.

tomob:
No problem. Here you are.

websockify/websocket.py
```diff
@@ -571,7 +571,7 @@ def _recvmsg(self):
                     msg = self._partial_msg
                     self._partial_msg = ''.decode("ascii")
                     return msg
-            elif frame["opcode"] == 0x2:
+            elif frame["opcode"] == 0x2 or frame["opcode"] == 0x1:
```
CendioOssman: This was an accident, I assume? :)

websockify/websocketproxy.py
```diff
@@ -418,6 +425,12 @@ def websockify_init():
             help="file of concatenated certificates of authorities trusted "
             "for validating clients (only effective with --verify-client). "
             "If omitted, system default list of CAs is used.")
+    parser.add_option("--ssl-version", choices=["tlsv1_1", "tlsv1_2"],
```
CendioOssman: We should be exhaustive here and allow older stuff in case people need it. And the default should probably be to let the `ssl` module keep its defaults.

(Python also supports TLS 1.3 now, so you might also want to add that)

tomob:
Regarding the fix commits: I don't why all test pass on my laptop, but fail on Travis. I don't know how Tox manages the python environments but that can something to do with it. Do you happen to have any advice on that?

websockify/websocketproxy.py
tomob: TLS 1.3 :+1: 
default: :+1:

But I'm a little hesitant to explicitly add options for versions that are considered insecure (TLS 1.0 and below). I think providing (potentially insecure) default + three secure options (TLS 1.1, 1.2 and 1.3) would make it both flexible and give users no option to make an unintentional mistake.

BTW, WebSocket has been standardized in 2011, while TLS 1.1 in 2006. I would expect that anyone using WebSockets has a software stack capable of running modern crypto.

CendioOssman:
There are no complaints about the tests right now, so I guess you solved it?

Anyway, merge the last two commits in to a single one and we should be good to merge.

tomob:
Yes, the tests are now fixed. It was an annoyance, not a real problem. I needed to push to see the Travis report instead of just running tests locally.

I've squashed the commits and rebased to get rid of merge conflict.

BTW, are you planning an "official" release anytime soon? It would help with managing dependencies.

CendioOssman:
Great. Thanks for your work.

As for a release, no one has really wanted to deal with that mess. The python code is probably in a good enough state, but everything else is of varying quality. So there is an open question of packaging.

Still, we could do a 0.9.0 "development release" I suppose.

tomob:
Thanks for the merge!

Dev release 0.9.0 would do for me. I'd appreciate that!

DirectXMan12:
@CendioOssman @kanaka @samhed I think it may be time to split the rest of the websockify implementations out into a separate repo, and/or mark them "use at your own risk".  We can then release *just* the python version.

kanaka:
@DirectXMan12 no objections. I am even okay with one repo per implementation if that makes sense. The main reason I could think of to keep the implementations together would be if we had some websockify tests that were generic and could test an arbitrary implementation for compatibility.

samhed:
@DirectXMan12 sounds good to me.

CendioOssman:
I would suggest:
 * **websockify** - Python code
 * **websockify-js** - Javascript (also popular, and works somewhat well)
 * **websockify-other** - The rest (poorly maintained and unknown status)

Sounds good?

kanaka:
@CendioOssman works for me.

PR #357, opened 2018-07-12
Port no longer included in token when using --host-token

josedpedroso:
Just a minor improvement to --host-token

### DIFF
--- a/websockify/sysloghandler.py
+++ b/websockify/sysloghandler.py
@@ -1,0 +1,118 @@
+import logging.handlers as handlers, socket, os, time
+
+
+class WebsockifySysLogHandler(handlers.SysLogHandler):
+    """
+    A handler class that sends proper Syslog-formatted messages,
+    as defined by RFC 5424.
+    """
+
+    _legacy_head_fmt = '<{pri}>{ident}[{pid}]: '
+    _rfc5424_head_fmt = '<{pri}>1 {timestamp} {hostname} {ident} {pid} - - '
+    _head_fmt = _rfc5424_head_fmt
+    _legacy = False
+    _timestamp_fmt = '%Y-%m-%dT%H:%M:%SZ'
+    _max_hostname = 255
+    _max_ident = 24 #safer for old daemons
+    _send_length = False
+    _tail = '\n'
+
+
+    ident = None
+
+
+    def __init__(self, address=('localhost', handlers.SYSLOG_UDP_PORT),
+                 facility=handlers.SysLogHandler.LOG_USER,
+                 socktype=None, ident=None, legacy=False):
+        """
+        Initialize a handler.
+
+        If address is specified as a string, a UNIX socket is used. To log to a
+        local syslogd, "WebsockifySysLogHandler(address="/dev/log")" can be
+        used. If facility is not specified, LOG_USER is used. If socktype is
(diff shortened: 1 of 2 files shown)


josedpedroso:
Take a look at https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SysLogHandler.emit

In python 3.3 and above, it's possible to use the ident attribute to identify the origin of a message, but there's no proper way to do it in earlier versions. Even then, the ident string is just prepended to the message, which is what I did with the syslog formatter.

Without that change, websockify's messages in syslog stand out for being the only ones without anything to identify where they came from.

I'll try to keep my commit messages more succint.

CendioOssman:
ident is an entirely separate field, so it isn't just prepended. Could you show me how your log lines look?

josedpedroso:
According to https://hg.python.org/cpython/rev/6baa90fa2b6d (2.15), ident really is just prepended to the message, but that's just the implementation of SysLogHandler in 3.3+. I'm not sure if that's really the way it should have been done.

I'll post a few messages from syslog with and without that change later today.

Now that I think about it, I should also try not setting any formatter at all for syslog, just to see what happens. But if many threads on stackoverflow are any indication, it'll be the same as using the regular log_formatter, ie, the messages will lack identification.

CendioOssman:
\*facepalm\*

So apparently Python implements some very basic, non-standardised version of the syslog protocol (which I assume is only accepted by the servers for historical reasons).

**ident** is a proper field in the protocol that allows you to separate different programs from each other. Putting it in the message part is a hack which does not achieve the same goal. If Python wants to do dumb things, then that's up to them. But I don't think we should play along.

If we want to fix this then have a look at RFC 5424 and do a subclass of `SysLogHandler` that actually sends the proper stuff. Or start using the Python `syslog` module. But that only works on Unix.

CendioOssman:
For reference, this is what a proper syslog log line looks like:
```
Jul 17 11:30:32 ossman systemd[1]: Starting dnf makecache...
```
It includes metadata like the hostname (`ossman`), the name of the process that sent, i.e. ident (`systemd`) and its pid (`1`). Only after the `:` is the actual message from the application. And these things are configurable and searchable, so we can never fake them by putting them in the message.

josedpedroso:
Alright, I haven't decided yet how I'm going to implement a proper fix for this, but I expect to have something over the next 2 weeks.

josedpedroso:
Hi. Please take a look at these changes. I had to keep support for the old syslog protocol as well, because I still use a syslogd that only parses that. It's not the default, though.

CendioOssman:
Nice. Looks good. Thanks for the work.

Could you do a new push without that last merge though? It just muddles thing.

CendioOssman:
In general you should just ignore what is happening upstream when you are working on a feature branch. The maintainers can handle the merging once things are ready.

To sort things out here you'll have to do:
```
$ git checkout master
$ git reset --hard d80c607
$ git cherry-pick ba702b2
```

If there are any conflicts, then you'll have to resolve those and run `git cherry-pick --continue`.

After that you can just run `git --force --push` and things should be updated here.

josedpedroso:
That did it. Thanks. I think it's all good now.

josedpedroso:
That did it. Thanks. I suppose you'll handle this conflict.

CendioOssman:
Great. Merged. Sorry for the delay.

PR #362, opened 2018-09-11
Remove all non-Python stuff from the tree

CendioOssman:
Suggested first commit after a split. Seems okay? If so I'll start setting up the other two repos.

One bit note though. Github doesn't have any way of moving issues. So it looks like we can't get them with us in any sane way to the other repos. My plan would be just to close them and users will have to open new ones.


CendioOssman:
> Are you actually intending to drop the wrap command (rebind) functionality? If so then you'll want to remove the corresponding python code from websockify/websocketproxy.py and update the man page (docs/websockify.1). And it would probably be best to remove the rebind files and python support code in a separate commit if dropping this support was intended.
> 
> If you weren't intending to drop that functionality (which is only currently supported in the python version BTW), then you'll want to restore Makefile, rebind and rebind.c.
> 

I wasn't aware it was hooked in to websockify like that. My original plan was to put it in "other". But if they are connected then I'll have to look at another approach.

> What's your plan for the client library and the examples (websock.js, wsirc, wstelnet)? Will they live in websockify-other? I would probably vote for those to stay here to provide example client code in the official websockify repo. But at a minimum we should document where to find them if they move.

I'd like to put them in `websockify-js`. I'd prefer for both clients and servers for each language to live together. I'd rather avoid having some clients be more "recommended" than others. This should be generic WebSocket stuff after all.

CendioOssman:
rebind is now kept. So I'll have a look at the other two repos and then I'll go ahead and merge this.

samhed:
Looks good, the PR's in the other repos have now been merged. Let's clean up:

- [ ] migrate issues to the javascript repo

- [ ] migrate issues to the 'other' repo

- [ ] merge this removal of the non-python stuff

asavah:
There are leftovers https://github.com/novnc/websockify/blob/master/setup.py#L21-L23 which prevent installing package via setup.py.
```
installing package data to build/bdist.x86_64-msib450ig-linux-gnu/egg
running install_data
creating build/bdist.x86_64-msib450ig-linux-gnu/egg/share
creating build/bdist.x86_64-msib450ig-linux-gnu/egg/share/websockify
creating build/bdist.x86_64-msib450ig-linux-gnu/egg/share/websockify/include
error: can't copy 'include/util.js': doesn't exist or not a regular file
```

samhed:
Good eye, removed those lines in 53004a0!

PR #371, opened 2018-11-28
removed symbolic link to includes for test cases to allow docker build to run on Windows

timothyclin:
converted the websockify include symbolic link in the test cases directory to static  files so docker build will success, Docker for Windows does not follow symbolic links properly with ADD/COPY commands due to NTFS limitation


CendioOssman:
I'm afraid this is not a good solution to your problem. This means we have duplicates of files, which means problems keeping things in sync.

A docker build should not have to run the tests to succeed, so just skip that stage if it cannot be run.

PR #384, opened 2019-03-07
Pyinstaller support

cowo78:
Added a [PyInstaller](https://www.pyinstaller.org/) .spec file to build a frozen distribution.
Also set "C" locale to limit required encodings modules.


CendioOssman:
Thank you for your contribution. This configuration file looks a bit large though, and difficult to maintain. Do we really need to have a large list of exclusion?

And we don't want to mess up the locale like that as we would like to respect the system settings.

cowo78:
Regarding size/maintenability: this is pretty much standard pyinstaller config, so I basically don't expect any changes to be needed (exept i.e. if websockify entry point changes). The exclusion list is of course optional: on python3.7/64bit leaving it out changes output size from 11MB to 14MB.

Regarding the locale: since no currencies are involved it basically means datetime. AFAIK logging in websockify has a custom formatter that only displays message so I don't see any changes happening.

However these are not standpoints to me so I'll stick to your decision.

CendioOssman:
Going from 11 MB to 14 MB doesn't seem like a big deal. Can we remove the entire exception list in that case?

cowo78:
Depends on how much flash you have on your device, but surely that's not a deal on common PC-based hardware.
I can comment the exclusion list, revert the setlocale and add some comments in the .spec on how to reduce output size.
Sounds good?

CendioOssman:
Sounds good. :+1:

cowo78:
Please have a look.
It may be interesting to distribute the self-contained binaries (think mainly on windows, though I don't even know if websockify works on the platform) along with other formats in a release cycle.

CendioOssman:
Looks good, thanks. Could you rebase everything in to a single commit?

>  It may be interesting to distribute the self-contained binaries (think mainly on windows, though I don't even know if websockify works on the platform) along with other formats in a release cycle.

Might be interesting, yes. I don't have much experience building self-contained python executables though.

PR #389, opened 2019-04-04
novnc redis backed token demo

PexMor:
This patch is extending the token backend (plugins) with redis in ```websockify/token_plugins.py``` file and then adds a demo for that purpose ```runNoVncWs.sh``` it actually pulls the novnc and patches ```novnc/app/ui.js``` to pass through the token from ```vnc.html?token=token``` to the ```websockify``` WS url.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,3 +2,5 @@ mox3
 nose
 jwcrypto;python_version>="2.7"
 enum34;python_version=="3.3"
+redis;python_version>="2.7"
+simplejson;python_version>="2.7"
--- a/websockify/auth_plugins.py
+++ b/websockify/auth_plugins.py
@@ -87,7 +87,7 @@ class ExpectOrigin(object):
         origin = headers.get('Origin', None)
         if origin is None or origin not in self.source:
             raise InvalidOriginError(expected=self.source, actual=origin)
-            
+
 class ClientCertCNAuth(object):
     """Verifies client by SSL certificate. Specify src as whitespace separated list of common names."""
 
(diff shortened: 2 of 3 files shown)


PexMor:
The test are failing on Python2.6, is that actually a problem?

PexMor:
Finally fixed the check by making it independent on any change to original code.

Would you consider including such proxy/firewall friendly demo in the main repo?

CendioOssman:
The redis class looks fine, but it is fairly small so I would prefer to have it in `token_plugin.py`. If you code defensively then it should be fine even on systems that lack the `redis` module.

The other changes do not look like something suited for upstream, I'm afraid.

PexMor:
Removed the demo part and merged into the ```auth_plugins.py```.  Hope that it could be marged into the upstream as a tiny piece.

PexMor:
Moved from auth_plugins.py to correct token_plugins.py.

Just an idea arises about those places, could the files rather be directories? What do you think?

CendioOssman:
> Moved from auth_plugins.py to correct token_plugins.py.

Great. Could you squash everything to a single commit and force update this branch?

> Just an idea arises about those places, could the files rather be directories? What do you think?

Sure, that sounds reasonable.

PexMor:
Squashed into one commit.

CendioOssman:
Great, thanks. :)

al3x609:
we need redis  dependence in setup.py, websockify master branch fail. :/

https://github.com/novnc/websockify/blob/70911c6d6ed7d05d2e8ffaba464efa18d045e582/setup.py#L32

```
  File "/mnt/noVNC/websockify/run", line 6, in <module>
    websockify.websocketproxy.websockify_init()
  File "/mnt/noVNC/websockify/websockify/websocketproxy.py", line 696, in websockify_init
    __import__(token_plugin_module)
  File "/mnt/noVNC/websockify/websockify/token_plugins.py", line 141, in <module>
    import redis
ImportError: No module named redis
```

PR #395, opened 2019-05-15
websocketproxy small glitch on win32

cowo78:
Fixed base class call to process_request() when usingThreadingMixIn


CendioOssman:
Oops. Thanks. :)

PR #398, opened 2019-05-28
Add a __main__.py file for invocation as a module

Hyask:
This file allows an easy invocation of websockify as a module, which is handy when running into shebang issues (like this: https://github.com/pypa/virtualenv/issues/596).
With this patch, `python -m websockify ...` has just the same behavior as `./run ...`.

### DIFF
--- a/run
+++ b/run
@@ -1,6 +1,7 @@
-#!/usr/bin/env python
+#!/usr/bin/env sh
 
-import websockify
+BASE_DIR="$(dirname "$(realpath "$0")")"
 
-if __name__ == '__main__':
-	websockify.websocketproxy.websockify_init()
+cd "$BASE_DIR"
+
+python -m websockify $@
--- a/websockify/__main__.py
+++ b/websockify/__main__.py
@@ -1,0 +1,4 @@
+import websockify
+
+if __name__ == '__main__':
+    websockify.websocketproxy.websockify_init()


Hyask:
It seems the CI setup is broken... :/

CendioOssman:
> It seems the CI setup is broken... :/

Yeah, Travis switched out their system so that Python 2.6 isn't available. I think it's possible to get back with some tweaks.

CendioOssman:
We're getting some duplication with this though. Could you change `run` so it just uses the new module way instead?

Hyask:
> We're getting some duplication with this though. Could you change `run` so it just uses the new module way instead?

I've changed `run` to be a simple shell script. Is that what you had in mind?

CendioOssman:
Yes, indeed. However it doesn't seem to quite work:

```
$ ~/devel/websockify/run 
/usr/bin/python: No module named websockify
```

It needs to change the working directory first as well so that Python finds the module.

Hyask:
> Yes, indeed. However it doesn't seem to quite work:
> 
> ```
> $ ~/devel/websockify/run 
> /usr/bin/python: No module named websockify
> ```
> 
> It needs to change the working directory first as well so that Python finds the module.

Oh, sorry, I did that quickly without many testings. It should be fixed now.

CendioOssman:
Nice, thanks.

run
```diff
@@ -1,6 +1,7 @@
...
+#!/usr/bin/env sh
 
-import websockify
+BASE_DIR="$(dirname "$(realpath "$0")")"
 
-if __name__ == '__main__':
-	websockify.websocketproxy.websockify_init()
+cd "$BASE_DIR"
+
+python -m websockify $@
```
andersk: This is missing double quotes around `"$@"`, which breaks invocation with arguments including spaces. Fixed in #405.

I recommend always using [ShellCheck](https://www.shellcheck.net/) to scan for common shell scripting errors like this.

Hyask: Thanks for the feedback, I'll keep that in mind.

PR #403, opened 2019-08-07
Include security headers required by security scan

serkankoyuncu:


CendioOssman:
Could you elaborate on when these are needed and why?

serkankoyuncu:
@CendioOssman
Qualys security scan shows some HTTP Security headers are missing in our server which runs noVnc in the cloud (Azure)
So we decided to clone websockify and add required HTTP headers in there instead of putting in behind a proxy something like nginx and we'd add required headers in there

I am not sure this is the best way to put them in there.

If we would have another option to add extra headers in websockify, then that would be great.

Here are the missing/required headers:
X-XSS-Protection HTTP Header missing on port 8443.
X-Content-Type-Options HTTP Header missing on port 8443.
Strict-Transport-Security HTTP Header missing on port 8443.

CendioOssman:
I'm afraid we can't really add things just because a tool just tool us to. Those headers look like something specific to your setup, so I don't think it is safe to add it here.

If you can figure out the purpose better then we can probably accept a patch that controls these headers via options. But I'm afraid this PR will not be accepted.

Issue #399, opened 2019-06-10
BUG-FIX: websockifyserver.py socket.getaddrinfo port need to be casted to str


caderize:
Dear WebSockify/noVNC Team

Testing noVNC/WebSockify against BaseTokenAPI I found that the connection failed
See: https://github.com/novnc/noVNC/issues/1250
 
The problem was "port" attribute of socket.getaddrinfo on websockifyserver.py (line: 461)

**From:**
`addrs = socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM,`

**To:**
`addrs = socket.getaddrinfo(host, int(port), 0, socket.SOCK_STREAM,`

**Patch** (diff -Naur websockifyserver.py.original websockifyserver.py) **:**

```
--- websockifyserver.py.original        2019-06-10 11:58:46.040671247 +0200
+++ websockifyserver.py 2019-06-10 11:57:53.632567186 +0200
@@ -458,7 +458,7 @@
             flags = flags | socket.AI_PASSIVE

         if not unix_socket:
-            addrs = socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM,
+            addrs = socket.getaddrinfo(host, int(port), 0, socket.SOCK_STREAM,
                     socket.IPPROTO_TCP, flags)
             if not addrs:
                 raise Exception("Could not resolve host '%s'" % host)
```

CendioOssman:
This is unfortunately incorrect, as port can be a name as well (e.g. `"ssh"` will look up port 22).

What platform are you seeing your issue on? I'm not getting any problem here.

caderize:
Dear Cendio Ossman,

you are right, 'port' is a const char*. So digging deeper I found the solution.
The problem is that using websockify with "--token-plugin=BaseTokenAPI" , port is then passed as an unicode string, but getaddrinfo() wants an ASCII string.

To reproduce the BUG:

1. Create a simple php page that echos "target_host:target_port", save it on the noVNC path
```
~$ cat /var/www/html/noVNC-1.1.0/token.php
<?php
echo "host_to_connect:5900"
?>
```

2.  Start websockify from the shell

`~$ /var/www/html/noVNC-1.1.0/utils/websockify/run --web /var/www/html/noVNC-1.1.0/ --token-plugin=BaseTokenAPI --token-source=http://myhost/noVNC-1.1.0/token.php?token=%s 8443
`

3. Open a browser to: http://myhost:8443/vnc.html?path=websockify?token=1111

[...]

PR #407, opened 2019-09-02
Coerce BaseTokenAPI port to int if numeric

bswinnerton:
This pull request builds on b46fab5 to further sanitize the port value that comes in when using the BaseTokenAPI plugin.

As far as I can tell, b46fab5 _only_ supports ascii based port values like `ssh` and not integer-based values like `22`. This results in identical failures to the ones described in novnc/websockify#399:

```
WebSocket server settings:
  - Listen on :80
  - No SSL/TLS support (no cert file)
  - proxying from :80 to targets generated by BaseTokenAPI
172.17.0.1 - - [02/Sep/2019 19:05:25] 172.17.0.1: Plain non-SSL (ws://) WebSocket connection
172.17.0.1 - - [02/Sep/2019 19:05:25] 172.17.0.1: Path: '/websockify?token=8bff5517-f620-461c-82df-6eb42163411b'
172.17.0.1 - - [02/Sep/2019 19:05:25] connecting to: 172.16.16.16:5910

172.17.0.1 - - [02/Sep/2019 19:05:25] Failed to connect to 172.16.16.16:5910
```

This commit makes two changes:
1. Strips any leading or trailing spaces from the port value
2. Checks to see if the ascii-encoded string is composed of digits, and if so, coerce the value to an `int`

### DIFF
--- a/websockify/token_plugins.py
+++ b/websockify/token_plugins.py
@@ -70,6 +70,9 @@ class BaseTokenAPI(BasePlugin):
     def process_result(self, resp):
         host, port = resp.text.split(':')
         port = port.encode('ascii','ignore')
+        port = str.strip(port)
+        if str.isdigit(port):
+            port = int(port)
         return [ host, port ]
 
     def lookup(self, token):


CendioOssman:
Hmm... giving port numbers as a string works fine for me. What crash are you seeing? And with what Python?

bswinnerton:
> What crash are you seeing?

The crash is identical to the one described in https://github.com/novnc/noVNC/issues/1250#issue-453572174: `Failed to connect to 172.16.16.16:5910`, as showcased above.

> what Python

```
$ python --version
Python 2.7.16
```

CendioOssman:
Works well for me. However note that most editors will put a newline in the file, which screws up the port. Could that be your issue?
In that case a simple `strip()` call might be a better fix here.

samhed:
No response.

PR #424, opened 2020-02-16
Remove str decode function from TokenRedis class

ryansen:
I'm using Python 2 version

image

if use str decode function, Websockify will be throw an error, the error code 1011.
remove the str decode function,  Websockify can connect the remote desktop.

### DIFF
--- a/websockify/token_plugins.py
+++ b/websockify/token_plugins.py
@@ -154,6 +154,6 @@ class TokenRedis(object):
         if stuff is None:
             return None
         else:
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm fairly sure this is needed for IDN. What Python version are you running exactly? And can you show us the error log?

tranquilryan:
@CendioOssman  Python 2.7.12
image

image

CendioOssman:
The issue was with the port, not the host. Fixed in 6d48b1507e7a0306dc60155ec7895be729b58f58.

PR #428, opened 2020-03-12
Add UnixDomainSocketDirectory token plugin

fengye87:
UnixDomainSocketDirectory token plugin can websockify all unix domain sockets under a specified directory. The token would be used as relative path to the specified path. It's very handy if QEMU is allocating VNC service on unix domain socket (with `vnc_auto_unix_socket` turned on).

Hope it helps.


fengye87:
@CendioOssman Added path check against escaping from the parent path, please review again.

PR #429, opened 2020-03-24
Add options --conn-plugin and --conn-source

jan-jockusch:
This PR adds new options for a python plugin which registers connection and disconnection events.

This is useful for external book-keeping of opened and closed connections. 

The options are fashioned in the same way as `--token-plugin` and `--token-source`.

### DIFF
--- a/websockify/conn_plugins.py
+++ b/websockify/conn_plugins.py
@@ -1,0 +1,36 @@
+class BasePlugin(object):
+    def __init__(self, src=None):
+        self.source = src
+
+    def connect(
+            self, host, port, use_ssl, unix_socket,
+            sockname, query):
+        """When a socket connection begins, this call receives as much
+        information as possible. Especially sockname is important, because
+        this allows "disconnect" to tell the connections apart."""
+        pass
+
+    def disconnect(
+            self, host, port, use_ssl, unix_socket,
+            sockname, query):
+        """This function is called with the exact same parameters as
+        "connect", but when the connection closes."""
+        pass
+
+
+class DebugPlugin(BasePlugin):
+    """Prints out event information for connections and disconnections."""
+
+    def connect(self, host, port, use_ssl, unix_socket, sockname, query):
+        print([
+            'conn_plugin.DebugPlugin connect',
+            host, port, use_ssl, unix_socket,
+            sockname, query,
+        ])
+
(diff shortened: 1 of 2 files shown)


jan-jockusch:
I don't really understand how to fix the Travis tests for Python2.6 and Python3.4.

Can anybody point me in the right direction so I can tidy up this Pull Request?

jan-jockusch:
Just removed a stray comment from the PR. I still don't understand how I can get Travis to work for all Python versions…

CendioOssman:
It seems Travis has changed something in their Python setup. We'll probably need to adjust things on our end.

jan-jockusch:
Okay, then I'll not worry about the two Python versions failing the tests (yet).

If there's anything else I should do to improve this PR let me know.

CendioOssman:
The code looks fine AFAICT, but I'm thinking that this is a bit overkill for just getting better logging. Have you looked at just improving the lines we currently print?

jan-jockusch:
Hi @CendioOssman ! Sorry for not answering sooner. I've accidentally deleted your message, that really shouldn't happen.

The reason for implementing this as a plugin is that it is for precise tracking of connections. We expect the connection to only proceed after it has been registered, and to be registered as "off" as soon as the connection is closed.

We also need port details, which is why we cannot mis-use or expand the auth plugin. A second plugin seemed like a much cleaner solution, because the connection details passed to the plugin are the same for connecting and disconnecting.

The (small) delay and possible instability introduced by log parsing are not acceptable in our use case, which is why I did not investigate that method further.

CendioOssman:
So this is infrastructure for a private plugin that you guys have? I'm not sure we want code in websockify that isn't used by our own code. Such things tend to rot and break.

@samhed, @kanaka, @DirectXMan12, what do you think?

samhed:
I don't think it's feasible to merge this.

jan-jockusch:
I won't press this. I made my point, and I can see why no-one cares for this addition, which is necessary for our application only.

We'll keep a fork and/or find another solution in the long term. 

Just one final question @CendioOssman: Would a PR containing expanded log lines pass more easily?

CendioOssman:
I'd say so. That has a more obvious general appeal, and doesn't really add any maintenance burden.

PR #437, opened 2020-07-28
Convert old print statement into function

sisou:
For Python 3 compatibility.

The current syntax is leading to a syntax error when used with Python 3.


CendioOssman:
Odd. I'm not getting a syntax error from Python 3 here. Still, it is incorrect so thanks for spotting this.

PR #438, opened 2020-07-28
Split token file by any whitespace after the colon, not just a space

sisou:
With the current parser logic, only tokens and servers that are separated by _exactly_ a colon and a space `: ` are detected as tokens. But when formatting one's token file with tabs, this breaks.

This commit changes the split characters to be a regular expression that matches all forms of whitespace, including spaces and tabs.

### DIFF
--- a/websockify/token_plugins.py
+++ b/websockify/token_plugins.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 import os
 import sys
+import re
 
 class BasePlugin(object):
     def __init__(self, src):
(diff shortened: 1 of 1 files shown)


CendioOssman:
Sounds sensible. Do you think you could add a unit test as well to make sure this behaviour is preserved in the future?

CendioOssman:
Ping @sisou .

sisou:
I am not familiar with Python testing. How does this library's tests work? I cannot find documentation about it. Can you give me a pointer on how to get started, what test to adapt maybe?

samhed:
There are some unittests for token in `tests/test_websocketproxy.py`, you could look at those.

PR #441, opened 2020-08-25
Use GitHub's recommended .gitignore file for Python projects

vuzdemav:
Taken from https://github.com/github/gitignore/blob/master/Python.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,7 @@
 target.cfg
 target.cfg.d
 .tox
+
+build/
+dist/
 *.egg-info


vuzdemav:
Actually all I wanted for myself was to add the `build/` and `dist/` directories created by `setup.py`, but I thought it is a good idea to use this template in case it is useful for others.

samhed:
I'm not sure I understand the purpose of having a generic .gitignore like that? I'd prefer to have a specific one for each project, to avoid confusions.

CendioOssman:
Yeah, I agree. This seems wildly excessive for our small project.

@vuzdemav, please adjust this PR with just the things that you were seeing and we'll add those.

vuzdemav:
I updated so that now it only adds `build/` and `dist/`

PR #455, opened 2020-10-21
fixed a bug in the function of send_frames.

codehasbug:
Today, I found a bug when I use the latest noVNC lib. Those are error pictures. When I use the another computer connect to host:port/websockify, it does not work well.
截图_2020-10-21_16-51-09

Due to a mouth ago, I had builded a test service. So, I check the code of websockify.
I discovered the $PRO_DIR/websockify/websockifyserver.py has a problem in function of send_frames.
In this part of the function, it send self.send_parts[0], but to pop the self.send_parts[end]. So, I did a small changed for it. Now it works well.

### DIFF
--- a/websockify/websockifyserver.py
+++ b/websockify/websockifyserver.py
@@ -149,10 +149,13 @@ class WebSockifyRequestHandler(WebSocketRequestHandlerMixIn, SimpleHTTPRequestHa
         while self.send_parts:
             # Send pending frames
             try:
-                self.request.sendmsg(self.send_parts.pop(0))
+                # send the first frame.
+                self.request.sendmsg(self.send_parts[0])
             except WebSocketWantWriteError:
                 self.print_traffic("<.")
                 return True
+            # pop the first frame. I think `self.send_parts.pop()` is incorrect. because the frame does not sended.
+            self.send_parts.pop(0)
             self.print_traffic("<")
 
         return False
--- a/websockify/websockifyserver.py
+++ b/websockify/websockifyserver.py
@@ -149,11 +149,10 @@ class WebSockifyRequestHandler(WebSocketRequestHandlerMixIn, SimpleHTTPRequestHa
         while self.send_parts:
             # Send pending frames
             try:
-                self.request.sendmsg(self.send_parts[0])
+                self.request.sendmsg(self.send_parts.pop(0))
             except WebSocketWantWriteError:
                 self.print_traffic("<.")
                 return True
-            self.send_parts.pop()
             self.print_traffic("<")
 
         return False


CendioOssman:
I'm afraid there must be something else that is broken as that is the correct way to use the API. Do you think you can have a look at `websocket.py` instead and see where things go wrong?

codehasbug:
I will check it later, but I have a question. Why do we write it this way(look at the picture)? Does the infrastructure have change? thank you. @CendioOssman
the diff

codehasbug:
And I think we just send the same element(the first) right now. If it is right, I am confused.
---原始邮件---
发件人:&quot;Pierre Ossman (Work account)&quot;<notifications@github.com&gt;
发送时间:2020年10月21日 星期三 下午6:52
收件人:&quot;novnc/websockify&quot;<websockify@noreply.github.com&gt;
抄送:&quot;JasonYang&quot;<1960228744@qq.com&gt;;&quot;Author&quot;<author@noreply.github.com&gt;
主题:Re: [novnc/websockify] fixed a bug in the function of send_frames. (#455)



I&#39;m&nbsp;afraid&nbsp;there&nbsp;must&nbsp;be&nbsp;something&nbsp;else&nbsp;that&nbsp;is&nbsp;broken&nbsp;as&nbsp;that&nbsp;is&nbsp;the&nbsp;correct&nbsp;way&nbsp;to&nbsp;use&nbsp;the&nbsp;API.&nbsp;Do&nbsp;you&nbsp;think&nbsp;you&nbsp;can&nbsp;have&nbsp;a&nbsp;look&nbsp;atwebsocket.pyinstead&nbsp;and&nbsp;see&nbsp;where&nbsp;things&nbsp;go&nbsp;wrong?

&mdash;
You&nbsp;are&nbsp;receiving&nbsp;this&nbsp;because&nbsp;you&nbsp;authored&nbsp;the&nbsp;thread.
Reply&nbsp;to&nbsp;this&nbsp;email&nbsp;directly,view&nbsp;it&nbsp;on&nbsp;GitHub,&nbsp;orunsubscribe.

CendioOssman:
To better match the SSLSocket and Socket APIs we mimic. Generally if something throws an exception or returns an error, then it didn't perform the desired action and you need to try again. So it's more reasonable to assume the data was not sent if an error was reported.

codehasbug:
> To better match the SSLSocket and Socket APIs we mimic. Generally if something throws an exception or returns an error, then it didn't perform the desired action and you need to try again. So it's more reasonable to assume the data was not sent if an error was reported.

Oh, I understand your mind. So, I did the pop later. But the issue is the sentence `self.send_parts.pop()`. It should be `self.send_parts.pop(0)`. For instance, we have a List [ 1, 2, 3, 4, 5 ] and we have to send them. If the sentence is `self.send_parts.pop()`, we will send 1 1 1 1 1. If the sentence is `self.send_parts.pop(0)`, we will send 1 2 3 4 5.

And I have checked `websocket.py` it just had some small changed, I cannot found any errors. In my opinion, It was just a careless mistake, to change `self.send_parts.pop()` to `self.send_parts.pop(0)`. it will be OK.

CendioOssman:
Ah. Good catch. Could I also bother you with trying to write a unit test so this issue doesn't reappear?

codehasbug:
> Ah. Good catch. Could I also bother you with trying to write a unit test so this issue doesn't reappear?

Sure. I write it later. If I finish it, I'll submit it.

CendioOssman:
Merged as 2b44803f83628656720537325606b55536d8108c. Thanks for you help here.

PR #472, opened 2021-01-27
Added exp claim for JWT token

jprietove:
Currently, JWT tokens are reusable and can be intercepted. By using Expiration Time registered claim, it is possible to let expire them.


CendioOssman:
Looks good. Is there also a start time we should check to protect against clock sync issues?

jprietove:
Yes, according to RFC 7519 section 4.1.5, there is 'nbf' Not Before
[https://tools.ietf.org/html/rfc7519#section-4.1](url)

PR #474, opened 2021-01-30
Simplify Windows instructions

patrakov:
The existing instructions reference a `noVNC Websocket Service.exe` file that either needs to be compiled or is [hard to find](https://github.com/downloads/kanaka/websockify/Websockify%20Windows%20Service.zip) because the downloads were not moved to the new project. It is also flagged by Firefox as something rarely downloaded (hinting at "possibly malicious"). For me, it didn't work on Windows 10 (starting the service takes forever). Also, it references `websockify.exe`, which is also hard to produce from a git checkout. The copies of `websockify.exe` available online (e.g. [this](https://gitee.com/forsrc/websockify-win)) have a crippling bug, they allow only one concurrent connection. And even then, the instructions fail to mention how to configure the firewall exception which would obviously be needed.

It does not need to be that painful, especially because there exists a much simpler way, involving the industry-standard `SrvAny.exe` tool. With that, I took the liberty of deleting the sources of the obviously unmaintained and now-unneeded Websockify-specific Windows service.


CendioOssman:
Looks good to me. Thanks for getting this sorted.

Issue #467, opened 2021-01-05
Redis not working?


skylar-byte:
The token plugin for Redis appears to not be working.

```
# websockify --token-plugin TokenRedis --token-source ./token_plugins.py 5900 --
Traceback (most recent call last):
  File "/opt/web_app/venv/bin/websockify", line 8, in <module>
    sys.exit(websockify_init())
  File "/opt/web_app/venv/lib/python3.8/site-packages/websockify/websocketproxy.py", line 699, in websockify_init
    opts.token_plugin = token_plugin_cls(opts.token_source)
  File "/opt/web_app/venv/lib/python3.8/site-packages/websockify/token_plugins.py", line 142, in __init__
    self._server, self._port = src.split(":")
ValueError: not enough values to unpack (expected 2, got 1)
```

The others appear to work fine.

```
# websockify --token-plugin BaseTokenAPI --token-source ./token_plugins.py 5900 --
WebSocket server settings:
  - Listen on :5900
  - No SSL/TLS support (no cert file)
  - proxying from :5900 to targets generated by BaseTokenAPI

# websockify --token-plugin JSONTokenApi --token-source ./token_plugins.py 5900 --
WebSocket server settings:
  - Listen on :5900
  - No SSL/TLS support (no cert file)
  - proxying from :5900 to targets generated by JSONTokenApi

# websockify --token-plugin JWTTokenApi --token-source ./token_plugins.py 5900 --
WebSocket server settings:
  - Listen on :5900
  - No SSL/TLS support (no cert file)
  - proxying from :5900 to targets generated by JWTTokenApi
```

I'm running the following versions:
```
# pip list
Package          Version
---------------- -------
redis            3.5.3  
websockify       0.9.0  
```

CendioOssman:
`--token-source` should specify your redis server in a `host:port` format. Please try that and it should work better.

skylar-byte:
@CendioOssman I've tried running the function manually from the tokens file and I was able to get a response which is weird.

```
# python3 ./test2.py 
None
['10.0.0.165', '5900']
# cat test2.py 
from websockify import token_plugins

print(token_plugins.TokenRedis.__init__(token_plugins.TokenRedis, "127.0.0.1:6379"))
print(token_plugins.TokenRedis.lookup(token_plugins.TokenRedis, "a91593ee7ed135104a96853fa2a7c042"))
```

CendioOssman:
[...]

PR #475, opened 2021-01-30
add docs and remove redundant ascii encoding

PexMor:
The fix for issue #467 also added some more docs about the testing and format of redis responses.

To-do: add better logging


CendioOssman:
Thanks for having a look. Please remove the old code though rather than commenting it out. We have git in case some one wants to find older versions. :)

PexMor:
Removed the extra comments and squashed.

CendioOssman:
Nice, thanks.

PR #511, opened 2022-04-12
Add tests for TokenRedis

javicacheiro:
Adding some basic tests for TokenRedis.


CendioOssman:
Looks good. 3.4 is failing, but that seems to be a general issue so I'll have a look at that in a separate PR.

PR #514, opened 2022-04-13
Use logging instead of directly printing messages to sys.stderr

javicacheiro:
The classes in token_plugins.py should use of the logging facility instead of printing directly to sys.stderr.

This way the logging configuration created in websockify_init() will also apply to these classes.

This does not affect the default behavior defined in websockify_init() that is to log to stderr, but it will have an important effect in the case that --log-file or --syslog options are used.

### DIFF
--- a/websockify/token_plugins.py
+++ b/websockify/token_plugins.py
@@ -1,7 +1,11 @@
+import logging
 import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


CendioOssman:
Looks good. Thanks!

PR #515, opened 2022-04-13
Apply configuration to the root logger

javicacheiro:
The global logging configuration should be applied to the root logger.

This way the configuration specified in the command line options will be propagated to all loggers.

This way all loggers will follow the configuration specified in the command line, and they will log to stderr, to a file or to syslog depending on the command line options given.


CendioOssman:
Sounds sensible. Thanks for the patch!

PR #516, opened 2022-04-13
Docker support

javicacheiro:
Adding support to run with docker/podman or any other compatible container runtime.

To simplify the upgrades, the container is built on the official python image, in this case python:3.6 but it would be very easy to move to a different python version.

The container is built using the latest official release. To generate the Dockerfile for a new release is as simple as to bump the VERSION variable.

I also kept the same dependencies as currently used in the CI process: numpy<1.17 redis simplejson jwcrypto.

Finally, the users can provide files using the standard /data volume (certificates, novnc).

If you create a dockerhub account, let's say it is named `novnc` then you can build the image with:
```
docker build -t novnc/websockify:0.10.0 .
docker push novnc/websockify:0.10.0
```

And then you can just update the README to point there, in the meanwhile I think it is useful to have the image accessible so users don't need to build it themselves again (since this precisely the main advantadge of using docker).

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,3 +168,36 @@ before running `python3 setup.py install`.
 
 Afterwards, websockify should be available in your path. Run
 `websockify --help` to confirm it's installed correctly.
+
+
+### Running with Docker/Podman
+You can also run websockify using Docker, Podman, Singularity, udocker or
+your favourite container runtime that support OCI container images.
+
+The entrypoint of the image is the `run` command.
+
+To build the image:
+```
+cd docker
+docker build -t novnc/websockify .
+```
+
+Once built you can just launch it with the same
+arguments you would give to the `run` command and taking care of
+assigning the port mappings:
+```
+docker run -it --rm -p <port>:<container_port> novnc/websockify <container_port> <run_arguments>
+```
+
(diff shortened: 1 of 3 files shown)


CendioOssman:
The other PR doesn't seem to resume, so let's go ahead here.

Your PR looks mostly fine. But it has the same issue as the other one; that it refers to an external repo. Please adjust the instructions so that shows how to get going with a local copy of websockify. The issue of having it in some repository is best dealt with separately.

javicacheiro:
Changed so now the image is build locally and there is no dependency on an external repo.

For convenience the name of the locally build image is "novnc/websockify" (you can name a locally built image as you want), so in the future when the project has a dockerhub account it would be just a matter of removing the docker build line from the instructions.

BTW, if you need help setting up the dockerhub repository and uploading the images just let me know.

CendioOssman (approved):
Nice, thanks!

The repo issue is more one of having yet another thing to keep track of and maintaining. Having it like this allows us to keep it a bit more "best effort" than having some CI job failing.

OliverEvans96:
Are there plans to publish the official image to dockerhub?

CendioOssman:
Not at the moment. I don't think anyone is opposed to it, but it's a bit of effort setting everything up.

PR #518, opened 2022-04-19
TokenRedis plugin improvements

javicacheiro:
Various improvements to the TokenRedis plugin maintaining backwards compatibility:
- Support for password protected redis servers
- Support to select the redis database to use
- Support both for json and text format tokens (json seems a bit of an overhead in most cases)
- Verification of the token decoding and return None in case of a decoding error
- Eliminate the dependency with simplejson module since the json module in the standard lib is more than enough for our basic usage of using the loads
- Additional tests to verify all the functionality

### DIFF
--- a/tests/test_token_plugins.py
+++ b/tests/test_token_plugins.py
@@ -203,3 +203,35 @@ class TokenRedisTestCase(unittest.TestCase):
         self.assertIsNotNone(result)
         self.assertEqual(result[0], 'remote_host')
         self.assertEqual(result[1], 'remote_port')
+
+    def test_src_only_host(self):
+        plugin = TokenRedis('127.0.0.1')
+
+        self.assertEqual(plugin._server, '127.0.0.1')
+        self.assertEqual(plugin._port, 6379)
+        self.assertEqual(plugin._db, 0)
+        self.assertEqual(plugin._password, None)
+
+    def test_src_with_host_port(self):
+        plugin = TokenRedis('127.0.0.1:1234')
+
+        self.assertEqual(plugin._server, '127.0.0.1')
+        self.assertEqual(plugin._port, 1234)
+        self.assertEqual(plugin._db, 0)
+        self.assertEqual(plugin._password, None)
+
+    def test_src_with_host_port_db(self):
+        plugin = TokenRedis('127.0.0.1:1234:2')
+
+        self.assertEqual(plugin._server, '127.0.0.1')
+        self.assertEqual(plugin._port, 1234)
+        self.assertEqual(plugin._db, 2)
+        self.assertEqual(plugin._password, None)
+
+    def test_src_with_host_port_db_pass(self):
+        plugin = TokenRedis('127.0.0.1:1234:2:verysecret')
+
+        self.assertEqual(plugin._server, '127.0.0.1')
+        self.assertEqual(plugin._port, 1234)
+        self.assertEqual(plugin._db, 2)
+        self.assertEqual(plugin._password, 'verysecret')
--- a/tests/test_token_plugins.py
+++ b/tests/test_token_plugins.py
@@ -204,6 +204,60 @@ class TokenRedisTestCase(unittest.TestCase):
         self.assertEqual(result[0], 'remote_host')
         self.assertEqual(result[1], 'remote_port')
 
+    @patch('redis.Redis')
+    def test_json_token_with_spaces(self, mock_redis):
+        plugin = TokenRedis('127.0.0.1:1234')
+
+        instance = mock_redis.return_value
+        instance.get.return_value = b' {"host": "remote_host:remote_port"} '
+
+        result = plugin.lookup('testhost')
+
+        instance.get.assert_called_once_with('testhost')
+        self.assertIsNotNone(result)
+        self.assertEqual(result[0], 'remote_host')
+        self.assertEqual(result[1], 'remote_port')
+
+    @patch('redis.Redis')
+    def test_text_token(self, mock_redis):
+        plugin = TokenRedis('127.0.0.1:1234')
+
+        instance = mock_redis.return_value
+        instance.get.return_value = b'remote_host:remote_port'
+
+        result = plugin.lookup('testhost')
+
+        instance.get.assert_called_once_with('testhost')
+        self.assertIsNotNone(result)
+        self.assertEqual(result[0], 'remote_host')
+        self.assertEqual(result[1], 'remote_port')
+
+    @patch('redis.Redis')
+    def test_text_token_with_spaces(self, mock_redis):
+        plugin = TokenRedis('127.0.0.1:1234')
+
+        instance = mock_redis.return_value
+        instance.get.return_value = b' remote_host:remote_port '
+
+        result = plugin.lookup('testhost')
+
+        instance.get.assert_called_once_with('testhost')
+        self.assertIsNotNone(result)
+        self.assertEqual(result[0], 'remote_host')
+        self.assertEqual(result[1], 'remote_port')
+
+    @patch('redis.Redis')
+    def test_invalid_token(self, mock_redis):
+        plugin = TokenRedis('127.0.0.1:1234')
+
+        instance = mock_redis.return_value
+        instance.get.return_value = b'{"host": "remote_host:remote_port"   '
+
+        result = plugin.lookup('testhost')
+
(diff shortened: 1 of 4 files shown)


javicacheiro:
Integration testing already performed against a password protected redis server.

CendioOssman:
Thanks for the contribution! But could you split this to several commits. One for each thing you'd like to fix.

javicacheiro:
Yes, of course, I was not sure the best way but this was also my initial thinking.

Do you want me to create different PR for each of them so then you can choose which ones to merge?

CendioOssman:
Let's start with everything here. Hopefully everything is okay. :)

javicacheiro:
Done!

websockify/websocketproxy.py
```diff
@@ -585,7 +587,8 @@ def websockify_init():
                                                  legacy=opts.legacy_syslog)
         syslog_handler.setLevel(logging.DEBUG)
         syslog_handler.setFormatter(log_formatter)
-        logger.addHandler(syslog_handler)
+        root = logging.getLogger()
+        root.addHandler(syslog_handler)
```
CendioOssman, line 591: Shouldn't this be part of #519?

websockify/token_plugins.py
```diff
@@ -176,6 +176,7 @@ def __init__(self, src):
             # import those ahead of time so we provide error earlier
             import redis
             self._server, self._port = src.split(":")
+            self._client = redis.Redis(host=self._server, port=self._port)
```
CendioOssman: I'm not sure whether this is safe to do. The server will fork for each request, so multiple requests might be using the same redis connection at the same time. That is likely not supported.

Have you examined this and checked if it is safe? A mere test run is not likely to expose this issue.

websockify/token_plugins.py
```diff
@@ -173,18 +190,36 @@ class TokenRedis():
...
+        try:
+            fields = src.split(":")
+            if len(fields) == 1:
+                self._server = fields[0]
+            elif len(fields) == 2:
+                self._server, self._port = fields
+            elif len(fields) == 3:
+                self._server, self._port, self._db = fields
+            elif len(fields) == 4:
+                self._server, self._port, self._db, self._password = fields
```
CendioOssman, line 224: Have you considered support if people want the default value for something "in the middle"? E.g. `host.example.com::dbname` would still use the default port.

CendioOssman (changes requested):
In general, I think this looks nice. Thanks!

There are some questions before proceeding, though. Please have a look.

websockify/websocketproxy.py
javicacheiro: Good question, actually this change is required for logging to work correctly with TokenRedis (and token_plugins in general).

In token_plugins.py we create a logger with:
```
    logger = logging.getLogger(__name__)
```
This is the standard way of using logging, in which the logger gets the __name__ of the module (in this case token_plugins), so in case someone is interested the logging level for this module can be tuned independently in the general configuration.

In the case of websocketproxy.py the logging instance used for logging is:
```
logging.getLogger(WebSocketProxy.log_prefix))
```
this is fine, and "WebSocketProxy.log_prefix" is used as the name of the logger, instead of using "__name__" to avoid it having the value "__main__".

The problem is that websocketproxy.py is also where the global logging is configured, so in order to enable also the logging of the other modules of the program we have to configure the root logger of the app. 

If we add the handler just to the "WebSocketProxy.log_prefix" logger then the messages from the other loggers in other modules of the app are lost.

This is the reason I had to update in websocketproxy.py all the places where the global logger is configured to use the root logger and not just the "WebSocketProxy.log_prefix" instance.

In PR519 what I try to address is something different, ie. increasing the logging level from INFO to DEBUG so we get all messages generated from the app (actually not many more, but I will address this in the other PR).

websockify/token_plugins.py
javicacheiro: Yes, this should be safe and more efficient sine we make use of the default redis connection pool, in other to reduce the overhead of creating/destroying connections constantly.

From redis-py documenation about [Connection Pools](https://github.com/redis/redis-py#connection-pools):
```
Behind the scenes, redis-py uses a connection pool to manage connections to a Redis server. By default, each Redis instance you create will in turn create its own connection pool.
```

Before, the code created a new redis client instance for each request with its associated connection pool, but then it did not make any use of the connection pool associated to the instance, because when a new connection came instead of reusing it will create a new instance.

For that reason, just moving the creation of the redis client instance to the contructor will make the trick (similarly to how it would be done for other databases).

CendioOssman:
Sorry for the delays. This is on my to-do list, but I'm a bit too busy with other things at the moment. I will get to this eventually, though. :)

javicacheiro:
No problem! In the meanwhile I have added the possibility you mentioned for empty options when using default values.

websockify/token_plugins.py
javicacheiro: Added support for this in the latest commit.

javicacheiro:
I did a rebase to integrate the fix for the new jwcrypto versions so tests do not break. I hope to not have created a mess.

javicacheiro:
Now is hard to see what are the commits specific to this PR.

Support for empty options is here:

https://github.com/novnc/websockify/pull/518/commits/8eeef0f6d77be3e6e8174a04a521a29b7a4c74cb

CendioOssman:
I'm afraid something did indeed go wrong when rebasing. Many other commits are suddenly included. Could you try again? `git rebase -i <upstream master branch>` and filtering the resulting list should do the right thing.

javicacheiro:
Fixed the problem with the rebase.

Everything looks fine now and  as, a side effect, we now have a rebase to the latest upstream. Really needed after so many months.

websockify/token_plugins.py
javicacheiro: I have been checking the redis-cli code.

When the Redis client is initialized without arguments, as we do, it creates a connection pool for that client instance. It is done here:
https://github.com/redis/redis-py/blob/3a121bef7bbc5bb5f07b119b0eef2f7527a38eda/redis/client.py#L1033

Then each time a command is executed it uses the current connection (if there is one active) or it gets one from the connection pool:
```
conn = self.connection or pool.get_connection(command_name, **options)
```
https://github.com/redis/redis-py/blob/3a121bef7bbc5bb5f07b119b0eef2f7527a38eda/redis/client.py#L1251

The documentation mentions that in case of using single_connection_client then no connection pool is used and it is not thread safe:
```
single_connection_client:
            if `True`, connection pool is not used. In that case `Redis`
            instance use is not thread safe.
```

If a new process is forked then it will have to instantiate its own TokenRedis and it will have its own connection pool, so you are right at the end in each forked process there will be a new connection pool. The advantage is that if this process does several gets then they will make use of the pool.

I have been running this version during the last months without issues, but we can do further testing. In that case I would propose to remove this part form this pull request and create a dedicated pull request just for this part.

Actually, I think the performance improvement will not be noticeable unless you have a very high number of requests because redis is pretty fast.

javicacheiro: Removed the connection pool part from current PR and I had to fix also the tests.

Since the connection pool part of the code was based on the changes in PR I will create a new PR when we have this part merged. What do you think?

CendioOssman: Sounds good.

CendioOssman (changes requested):
Just a small commit history adjustment and I think it's ready for merging.

Thanks for your work!

tests/test_token_plugins.py
```diff
@@ -266,8 +266,6 @@ def test_src_only_host(self, mock_redis):
         self.assertEqual(plugin._port, 6379)
         self.assertEqual(plugin._db, 0)
         self.assertEqual(plugin._password, None)
-        mock_redis.assert_called_once_with(host='127.0.0.1', port=6379,
-                                           db=0, password=None)
```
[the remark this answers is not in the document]
javicacheiro: Done!

javicacheiro:
> Just a small commit history adjustment and I think it's ready for merging.
> 
> Thanks for your work!

Happy New Year!

Sorry for the delay, the history is now adjusted and it looks much better without the back and forth.

When this PR is merged I will create the new one with the connection pool part based on this code.

CendioOssman (approved):
Looks good. Thanks!

PR #519, opened 2022-05-03
When using the --verbose option set root log level to DEBUG

javicacheiro:
When using the --verbose/-v option set root log level to DEBUG

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -413,15 +413,13 @@ def select_ssl_version(version):
 
 def websockify_init():
     # Setup basic logging to stderr.
-    logger = logging.getLogger(WebSocketProxy.log_prefix)
-    logger.propagate = False
-    logger.setLevel(logging.INFO)
     stderr_handler = logging.StreamHandler()
     stderr_handler.setLevel(logging.DEBUG)
     log_formatter = logging.Formatter("%(message)s")
     stderr_handler.setFormatter(log_formatter)
     root = logging.getLogger()
     root.addHandler(stderr_handler)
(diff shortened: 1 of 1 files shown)


samhed:
:thinking: is this common practice?

javicacheiro:
It depends on the external modules that you use and how verbose they are.

In the case of websockify it does not increase the logging messages much and it is quite useful for debugging so I think it is quite convenient and it has no obvious drawbacks.

CendioOssman:
We seem to have already crossed the bridge of mixing up "debug" and "verbose", so that should not be a blocker here.

But I'm confused about what this PR actually does. What log messages are appearing at different levels of the hierarchy?

javicacheiro:
Sorry, I forgot to include in the previous comment the output we would get without the "--verbose" option:
```
TokenRedis backend initilized (3.3.3.3:6379)
WebSocket server settings:
  - Listen on 1.1.1.1:9444
  - Web server. Web root: /root/websockify/standalone/novnc/noVNC-1.3.0
  - SSL/TLS support
  - Deny non-SSL/TLS connections
  - proxying from 1.1.1.1:9444 to targets generated by TokenRedis
Resolving token 'TKNXXXX'
2.2.2.2 - - [16/May/2022 17:40:52] 2.2.2.2: SSL/TLS (wss://) WebSocket connection
2.2.2.2 - - [16/May/2022 17:40:52] 2.2.2.2: Path: '/websockify?token=TKNXXXX'
2.2.2.2 - - [16/May/2022 17:40:52] connecting to: 4.4.4.4:5901
```

samhed (approved):
Looks good to me

PR #521, opened 2022-05-25
Tests break with jwcrypto>=1.3

javicacheiro:
Newly realeased jwcryto versions 1.3 and 1.3.1 (both released this month) breaks the JWTTokenApi class and also the testing suite.

The reason is that the JWTTokenApi class uses a no longer supported way of creating a key:
```
key = jwt.JWK()
```

Actually instead of the **jwt** module it should use the **jwk** module for JWK():
```
key = jwk.JWK()
```

From 1.3 the jwt module no longer exposes JWK so the only option is to use the correct module.

### DIFF
--- a/tests/test_token_plugins.py
+++ b/tests/test_token_plugins.py
@@ -4,7 +4,7 @@
 
 import unittest
 from unittest.mock import patch, mock_open, MagicMock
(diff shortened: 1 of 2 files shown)


samhed:
Thanks!

PR #525, opened 2022-08-11
Drain queues before closing connections

### LINKED ISSUE #513: Data loss while transmitting large amounts of data
I think this is related to or at least very similar to #63 and #36.

We've discovered an issue when transmitting large amounts of data (in our case from target to client).

When the target completed the transmission of the data, it can happen that the target closes the connection while websockify still has data in the client-side buffer.
Then, websockify closes also the client-side connection, which leads to a loss of data.

We see in a packet capture that websockify receives all of the data, but not all of the data is written to the websocket.
The last message in the websocket is "Target closed".

We've debugged further and realized that the moment websockify closes the client connection len(cqueue) is != 0, so there's still data in the client-side buffer which has to be sent.

I assume that this is unwanted behavior.

We've implemented a workaround by checking if there's still data in the buffer before closing the client-side socket.

While I'm unsure if there are any side-effects of doing this, it seems to solve our problem.

            if target in ins:
                # Receive target data, encode it and queue for client
                buf = target.recv(self.buffer_size)
                if len(buf) == 0 and len(cqueue) == 0:
                    if self.verbose:
                        self.log_message("%s:%s: Target closed connection",
                                self.server.target_host, self.server.target_port)
                    raise self.CClose(1000, "Target closed")

                if len(buf) != 0:
                    cqueue.append(buf)
                    self.print_traffic("{")

msnatepg:
This PR fixes #513.

Instead of looping until the queues are empty (as suggested by me in the issue) it drains the queues in the same loop pass.

Code could be refactored to reduce duplication but the potential for improvement here is not that big.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -222,6 +222,18 @@ Traffic Legend:
                 tqueue.extend(bufs)
 
                 if closed:
+
+                    while (len(tqueue) != 0):
+                        # Send queued client data to the target
+                        dat = tqueue.pop(0)
+                        sent = target.send(dat)
+                        if sent == len(dat):
+                            self.print_traffic(">")
+                        else:
+                            # requeue the remaining data
+                            tqueue.insert(0, dat[sent:])
+                            self.print_traffic(".>")
+
                     # TODO: What about blocking on client socket?
                     if self.verbose:
                         self.log_message("%s:%s: Client closed connection",
@@ -245,6 +257,16 @@ Traffic Legend:
                 # Receive target data, encode it and queue for client
                 buf = target.recv(self.buffer_size)
                 if len(buf) == 0:
+
+                    # Target socket closed, flushing queues and closing client-side websocket
+                    # Send queued target data to the client
+                    if len(cqueue) != 0:
+                        c_pend = True
+                        while(c_pend):
+                            c_pend = self.send_frames(cqueue)
+
+                        cqueue = []
+
                     if self.verbose:
                         self.log_message("%s:%s: Target closed connection",
                                 self.server.target_host, self.server.target_port)


websockify/websocketproxy.py
```diff
@@ -222,6 +222,18 @@ def do_proxy(self, target):
                 tqueue.extend(bufs)
 
                 if closed:
+
+                    while (len(tqueue) != 0):
+                        # Send queued client data to the target
+                        dat = tqueue.pop(0)
+                        sent = target.send(dat)
```
CendioOssman, line 229: Ideally, this should be a `select()` loop, but since the socket seems to be blocking this should work anyway.

websockify/websocketproxy.py
```diff
@@ -245,13 +257,19 @@ def do_proxy(self, target):
                 # Receive target data, encode it and queue for client
                 buf = target.recv(self.buffer_size)
                 if len(buf) == 0:
+
+                    # Send queued target data to the client
+                    c_pend = self.send_frames(cqueue)
+                    cqueue = []
```
CendioOssman, line 263: This does nothing as far as I can tell. If there was any data in `cqueue`, then `send_frames()` would have already been called.

The buffering issue is deeper, where you need to keep calling `send_frames()` until `c_pend` is `False`.

I'm concerned what tests you were doing, though. I would have assumed you would have seen a lack of flush here?

websockify/websocketproxy.py
```diff
@@ -245,13 +257,19 @@ def do_proxy(self, target):
...
                     if self.verbose:
                         self.log_message("%s:%s: Target closed connection",
                                 self.server.target_host, self.server.target_port)
                     raise self.CClose(1000, "Target closed")
 
-                cqueue.append(buf)
-                self.print_traffic("{")
+                if len(buf) > 0:
+                    cqueue.append(buf)
+                    self.print_traffic("{")
```
CendioOssman, line 272: This change doesn't look meaningful. How can it reach this without `len(buf) > 0`?

CendioOssman (changes requested):
I think the principle is okay, so we just need some tweaks here.

websockify/websocketproxy.py
```diff
@@ -212,7 +212,10 @@ def do_proxy(self, target):
 
             if self.request in outs:
                 # Send queued target data to the client
-                c_pend = self.send_frames(cqueue)
+                # As do-until loops are not available in python, settings c_pend to true
+                c_pend = True
+                while(c_pend):
+                    c_pend = self.send_frames(cqueue)
```
CendioOssman, line 218: I'm afraid this makes things less responsive, as it will now block here until it can send everything to the browser. That means it will pause sending things in the other direction.

websockify/websocketproxy.py
```diff
@@ -35,15 +35,15 @@ class ProxyRequestHandler(websockifyserver.WebSockifyRequestHandler):
     <  - Client send
     <. - Client send partial
 """
-
+
```
CendioOssman, line 38: We don't want any trailing whitespace.

websockify/websocketproxy.py
```diff
@@ -193,8 +192,9 @@ def do_proxy(self, target):
                     self.send_ping()
 
             if tqueue: wlist.append(target)
-            if cqueue or c_pend: wlist.append(self.request)
+            if cqueue: wlist.append(self.request)
             try:
+                # blocks until there is data from one side
```
CendioOssman, line 197: Or it's possible to send more data.

I'm not sure if this is very useful. An understanding of `select()` is probably required anyway before touching this code.

websockify/websocketproxy.py
```diff
@@ -257,19 +260,24 @@ def do_proxy(self, target):
                 # Receive target data, encode it and queue for client
                 buf = target.recv(self.buffer_size)
                 if len(buf) == 0:
-
+                    # Target socket closed, flushing queues and closing client-side websocket
                     # Send queued target data to the client
-                    c_pend = self.send_frames(cqueue)
-                    cqueue = []
+                    while (len(cqueue) != 0):
+                        # As do-until loops are not available in python, settings c_pend to true
```
CendioOssman, line 266: I don't think we need programming language commentary. :)

You should also be able to do:

```python3
while (self.send_frames(cqueue)):
    pass
```

websockify/websocketproxy.py
```diff
@@ -257,19 +260,24 @@ def do_proxy(self, target):
                 # Receive target data, encode it and queue for client
                 buf = target.recv(self.buffer_size)
                 if len(buf) == 0:
-
+                    # Target socket closed, flushing queues and closing client-side websocket
                     # Send queued target data to the client
-                    c_pend = self.send_frames(cqueue)
-                    cqueue = []
+                    while (len(cqueue) != 0):
```
CendioOssman, line 265: `if` should be better here, as it will never loop.

websockify/websocketproxy.py
```diff
@@ -438,17 +446,12 @@ def websockify_init():
     stderr_handler.setLevel(logging.DEBUG)
     log_formatter = logging.Formatter("%(message)s")
     stderr_handler.setFormatter(log_formatter)
-    root = logging.getLogger()
-    root.addHandler(stderr_handler)
+    logger.addHandler(stderr_handler)
```
CendioOssman, line 449: This seems unrelated.

CendioOssman (changes requested):
Sorry for the late reply. You should have some more comments now.

CendioOssman (approved):
Looks good. Do you think you could squash the commits together? Then I'll go ahead and merge this.

msnatepg:
Commit 9a554fc8b398990f8a621e8673a2e261a31bcdee "changes" was not intended to be pushed to GitHub, I'm sorry for this.
This was for an internal test which was not based on the current master branch, therefore there are other changes as well.

As you already have discovered, I have reverted this commit and commited only the necessary changes.
I will squash these commits together so that we can merge this.

Thank you for your support!

PR #536, opened 2022-10-18
expose --file-only option to disable dir listing

msays2000:
some security scanning tools complain that exposing a directory for directory listing is a security violation. A option already exists to disable dir listing but it is not exposed to be used.

This PR simply exposes that option.


msays2000:
@CendioOssman please review this to expose a flag, this has widely been asked in many online forums example:
https://groups.google.com/g/novnc/c/XVnJ7A0Kd5U

https://github.com/novnc/noVNC/issues/294

CendioOssman:
Great, thanks!

Issue #539, opened 2022-11-08
Support for Listening to unix socket [+ git patch for v0.10.0]


shiomax:
I´m using websockify and two other services behind NGINX in a docker container. Listening to an IP in this setup is not ideal. Allocates an extra port and a Unix socket should also be slightly faster.

I´ve already implemented a patch for websockify v0.10.0.

This would add two more options

```
  --unix-listen=FILE    listen to unix socket
  --unix-listen-mode=LISTEN_SOCK_MODE
                        specify mode for unix socket (defaults to 0600)
```

I used 0600 as the default access modifier as that's what tigervnc seems to default to.

Can work on a merge request if this seems fine to you.

You can apply the patch against either the git tag v0.10.0 or the tar releases (but I have not tested if this works with the latest commit)
Either with
- Git ```git apply <patchfile>``` (I hope... I never used this command)
- Linux patch command ```patch --directory /path/to/websockify -p1 </path/to/patch```

```
diff --git a/websockify/websocketproxy.py b/websockify/websocketproxy.py
index 09d7882..16199f0 100644
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -11,7 +11,7 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import signal, socket, optparse, time, os, sys, subprocess, logging, errno, ssl
+import signal, socket, optparse, time, os, sys, subprocess, logging, errno, ssl, stat
 from socketserver import ThreadingMixIn
 from http.server import HTTPServer
 
@@ -112,7 +112,9 @@ Traffic Legend:
                              self.server.target_host, self.server.target_port, e)
             raise self.CClose(1011, "Failed to connect to downstream server")
 
-        self.request.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, 1)
+        # Option unavailable when listening to unix socket
+        if not self.server.listen_sock:
+            self.request.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, 1)
         if not self.server.wrap_cmd and not self.server.unix_target:
             tsock.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, 1)
 
@@ -467,6 +469,12 @@ def websockify_init():
     parser.add_option("--ssl-ciphers", action="store",
             help="list of ciphers allowed for connection. For a list of "
             "supported ciphers run `openssl ciphers`")
+    parser.add_option("--unix-listen",
[...]

PR #541, opened 2022-11-08
Add option to listen to unix socket

shiomax:
https://github.com/novnc/websockify/issues/539

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -11,7 +11,7 @@ as taken from http://docs.python.org/dev/library/ssl.html#certificates
 
 '''
 
-import signal, socket, optparse, time, os, sys, subprocess, logging, errno, ssl
+import signal, socket, optparse, time, os, sys, subprocess, logging, errno, ssl, stat
 from socketserver import ThreadingMixIn
 from http.server import HTTPServer
 
@@ -112,7 +112,9 @@ Traffic Legend:
                              self.server.target_host, self.server.target_port, e)
             raise self.CClose(1011, "Failed to connect to downstream server")
 
-        self.request.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, 1)
+        # Option unavailable when listening to unix socket
+        if not self.server.unix_listen:
+            self.request.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, 1)
         if not self.server.wrap_cmd and not self.server.unix_target:
             tsock.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, 1)
 
@@ -467,6 +469,10 @@ def websockify_init():
     parser.add_option("--ssl-ciphers", action="store",
             help="list of ciphers allowed for connection. For a list of "
             "supported ciphers run `openssl ciphers`")
+    parser.add_option("--unix-listen",
+            help="listen to unix socket", metavar="FILE", default=None)
+    parser.add_option("--unix-listen-mode", default=None,
+            help="specify mode for unix socket (defaults to 0600)")
     parser.add_option("--unix-target",
             help="connect to unix socket target", metavar="FILE")
     parser.add_option("--inetd",
@@ -617,6 +623,16 @@ def websockify_init():
 
     if opts.inetd:
         opts.listen_fd = sys.stdin.fileno()
+    elif opts.unix_listen:
+        if opts.unix_listen_mode:
+            try:
+                # Parse octal notation (like 750)
+                opts.unix_listen_mode = int(opts.unix_listen_mode, 8)
+            except ValueError:
+                parser.error("Error parsing listen unix socket mode")
+        else:
+            # Default to 0600 (Owner Read/Write)
+            opts.unix_listen_mode = stat.S_IREAD | stat.S_IWRITE
     else:
         if len(args) < 1:
             parser.error("Too few arguments")
--- a/websockify/websockifyserver.py
+++ b/websockify/websockifyserver.py
@@ -325,37 +325,40 @@ class WebSockifyServer():
             file_only=False,
             run_once=False, timeout=0, idle_timeout=0, traffic=False,
             tcp_keepalive=True, tcp_keepcnt=None, tcp_keepidle=None,
-            tcp_keepintvl=None, ssl_ciphers=None, ssl_options=0):
+            tcp_keepintvl=None, ssl_ciphers=None, ssl_options=0,
+            unix_listen=None, unix_listen_mode=None):
 
         # settings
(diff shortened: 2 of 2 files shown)


websockify/websockifyserver.py
```diff
@@ -700,6 +705,17 @@ def start_server(self):
 
         if self.listen_fd != None:
             lsock = socket.fromfd(self.listen_fd, socket.AF_INET, socket.SOCK_STREAM)
+        elif self.listen_sock != None:
+            # Make sure the socket does not already exist
+            try:
+                os.unlink(self.listen_sock)
+            except OSError:
+                if os.path.exists(self.listen_sock):
+                    raise
```
CendioOssman, line 714: You can just do this instead:

```python3
except FileNotFoundError:
    pass
```

websockify/websockifyserver.py
```diff
@@ -700,6 +705,17 @@ def start_server(self):
...
+            lsock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+            lsock.bind(self.listen_sock)
+            os.chmod(self.listen_sock, self.listen_sock_mode)
```
CendioOssman, line 717: I wonder if this is race:y...

The safe way is to use umask. I.e.:

```python3
oldmask = os.umask(0777 ^ self.listen_sock_mode)
try:
    lsock.bind(self.listen_sock)
finally:
    os.umask(oldmask)
```

websockify/websockifyserver.py
```diff
@@ -700,6 +705,17 @@ def start_server(self):
...
+            lsock.listen(100)
         else:
             lsock = self.socket(self.listen_host, self.listen_port, False,
```
CendioOssman, line 720: I think you should probably move the socket setup in to this function. It already handles Unix sockets, just not listening ones.

websockify/websocketproxy.py
```diff
@@ -492,6 +494,12 @@ def websockify_init():
     parser.add_option("--ssl-ciphers", action="store",
             help="list of ciphers allowed for connection. For a list of "
             "supported ciphers run `openssl ciphers`")
+    parser.add_option("--unix-listen",
+            dest="listen_sock",
```
CendioOssman, line 498: We've avoided using `dest` for pretty much everything else, so let's try to keep the same pattern for new stuff.

CendioOssman (changes requested):

CendioOssman:
Did you check if there are some unit tests that could be added for this change?

websockify/websockifyserver.py
shiomax: Not 100% sure that there isn´t a race condition, so far it worked for me, but that does not say too much.

With ```chmod``` you`d be sure you´re only changing permissions on that one file, while with ```umask``` there might be ways you end up changing permissions on other files that are created simultaneously.

Also if there where a race condition, would not this suffer from the same issue? Just instead of ```chmod``` failing it would change back to the old umask before the file was created?

And, whats this for? ```0777 ^```

websockify/websockifyserver.py
shiomax: Sure.

shiomax:
> Did you check if there are some unit tests that could be added for this change?

Not haven´t checked how to unit test this.

websockify/websockifyserver.py
shiomax: Can do, I´ve seen the method was easier to put it in there at first.

Not entierly sure about the ```lsock.listen(100)```. I took the 100 from somewhere else in the code. But unsure why it would be needed. Probably not necessary to pass anything to the constructor here.

websockify/websocketproxy.py
shiomax: Would ```--listen-unix-sock``` / ```listen_unix_sock``` and ```--listen-unix-sock-mode``` / ```listen_unix_sock_mode``` be fine instead? Or remove the ```-sock``` to make it shorter?

Personally, I find the unix part to be more telling because there are many other socket types one can listen to but with unix listen I would generally default to thinking thats a unix socket automatically. Though, maybe that´s just me.

shiomax: Did remove ```dest``` now. Though, the reason I did pick the names I did is because there already is ```--unix-target```, so made some sence to call it ```--unix-listen``` in the CLI option to me. However, the variables in the code for listen all start with listen ```listen_host``` / ```listen_port``` / ```listen_fd``` etc.

For the ```--inetd``` option ```dest``` is not used but it´s rewritten later to another variable.

```
if opts.inetd:
   opts.listen_fd = sys.stdin.fileno()
```

Should I do that or just leave it as is? (with the variables called ```unix_target``` / ```unix_target_mode```)

websockify/websockifyserver.py
CendioOssman: > With chmod you`d be sure you´re only changing permissions on that one file (while being unfamiliar with the project and only looking at that one line of code), while with umask there might be ways you end up changing permissions on other files that are created simultaneously. Though, if you´re sure about that change can do ofc.

We don't have threads (at least not at this point), so the `finally` should be enough to make sure we don't mess with any other files.

> Also if there where a race condition, would not this suffer from the same issue? Just instead of chmod failing it would change back to the old umask before the file was created?

The race isn't about `chmod()` failing, but rather about someone opening the socket before `chmod()` has been run. This new code is safe since there is no point where the socket exists with the wrong permissions.

> And, whats this for? 0777 ^

`umask()` is inverted in that it states the bits you *don't want*, rather than the bits you *want*. Did you check if you go the expected result with the new code?

websockify/websockifyserver.py
CendioOssman: You got that 100 from the socket setup function. So things would definitely clearer with all `listen()` calls in the same place. :)

websockify/websockifyserver.py
maximal096: Sorry for the long time without any response.

You´re right about that. It´s indeed needed.

I did make that change too. Besides the lack of tests it now looks good? Gonna have to take a look at the existing tests and see whats tested there. Seems hard to test proxies. It probably would be enough if I just manage to validate the new settings work and send any packet at all over the socket file. As all the rest should (tm) work the same.

shiomax: Accidentally posted this with the wrong account. Did not know I was logged in with another account. Mostly using gitlab myself. Should really delete the other one not much point in it.

shiomax:
> 

I did do that now. Still, have a branch with the individual commits if anything went wrong can try again. Though, the resulting commit looks fine to me. Also tested if it works and it does.

CendioOssman (approved):
Looks good. Thank you for your contribution!

PR #542, opened 2022-11-17
Preserve existing LD_PRELOAD, prepend rebinder 

agt:
Insert rebinder at the head of the (possibly empty) LD_PRELOAD pathlist,
required for use cases relying on other preloaded libraries e.g. nss_wrapper.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -325,8 +325,11 @@ class WebSocketProxy(websockifyserver.WebSockifyServer):
             self.target_port = sock.getsockname()[1]
             sock.close()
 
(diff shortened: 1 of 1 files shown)


agt:
A bit of additional context - we LD_PRELOAD "nss_wrapper" to insert meaningful passwd/group entries into unprivileged containers which run under user-specific but arbitrary uids, similar to [Openshift's approach](https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines) , and need this configuration to propagate to websockified child processes.

CendioOssman:
Looks good, thanks.

PR #546, opened 2023-01-07
Update README.md

donbright:
how to add ssl exceptions for WSS urls, prevent the SSL Invalid Certificate error in Websockify


CendioOssman (approved):
Looks like a good addition. Thanks!

PR #549, opened 2023-02-26
Added traffic plugin support

gschaden:
I wanted to use the this project together with noVNC to allow our customers/qa to connect to vnc devices(linux) on  our private networks. I do not want them to know/enter passwords, but the devices have passwords configured. The traffic plugin system allows me to intercept the authentication and handle it differently. 
The example implementation takes a password provided by the JSONTokenApi 

`
--traffic-plugin websockify.plugins.VncTokenAuthenticationTrafficPlugin \
--token-plugin JSONTokenApi \
--token-source https://someapie/verify?token=%s
`
The API should respond with
`
{
  "host": "localhost",
  "port": 5902,
  "password": "vncpassword"
}
`

### DIFF
--- a/websockify/plugins.py
+++ b/websockify/plugins.py
@@ -1,0 +1,47 @@
+from websockify.traffic_plugins import BasePlugin
+from websockify import pyDes
+
+class RFBDes(pyDes.des):
+    def setKey(self, key):
+        """RFB protocol for authentication requires client to encrypt
+           challenge sent by server with password using DES method. However,
+           bits in each byte of the password are put in reverse order before
+           using it as encryption key."""
+        newkey = []
+        for ki in range(len(key)):
+            bsrc = ord(key[ki])
+            btgt = 0
+            for i in range(8):
+                if bsrc & (1 << i):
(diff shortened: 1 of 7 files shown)


websockify/plugins.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def from_client(self, s):
+        self.client_packet_count += 1
+        if self.client_packet_count == 2:
+            return None
+        return s
+
+    def from_target(self, s):
+        self.target_packet_count += 1
+        if self.target_packet_count == 2 and b"\x02" in s[1:]: # check if password is supported
```
CendioOssman, line 36: I'm afraid this is a bit fragile. There is no guarantee that the stream will be split in packages in a specific way.

Some rudimentary protocol handling will be needed here.

Have you considered a proxy model, instead of this filtering model? It might be a better fit.

CendioOssman (changes requested):
I can understand the use case here, but I'm a bit cautious about where we draw the line of what should be included in this project.

So far, we've stayed very agnostic about what kind of traffic flows through websockify.

PR #567, opened 2024-01-11
feat: TokenFileName maybe a better way

shifengbin:
TokenFileName maybe a better way


shifengbin:
in order to facilitate debugging and retrieval

PR #570, opened 2024-02-16
Make wrap_cmd work as a simple run when using unix sockets

dylex:
If unix-target is specified, don't use rebind (fixes a crash when accessing target_port).  If unix-listen is specified, there is no port argument to skip.

Currently if you specify a wrap_cmd along with a unix_listen socket or a unix_target, things break because of the lack of port.  Admittedly there is little reason to use wrap_cmd in this case because you could just run it separately, but it's still a convenient to just run the command.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -302,7 +302,7 @@ class WebSocketProxy(websockifyserver.WebSockifyServer):
         # Last 3 timestamps command was run
         self.wrap_times    = [0, 0, 0]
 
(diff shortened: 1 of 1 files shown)


CendioOssman:
Maybe I'm missing something, but if I understand you correctly, websockify crashes if you try to combine wrapping with a unix socket target?

That seems like a bug we should fix in that case, instead of disabling wrapping. That will just confuse the user, as websockify will not do what they asked.

PR #580, opened 2024-06-03
drop python<=3.7 support

kloczek:
Filter all code over `pyupgrade --py38-plus`.


CendioOssman:
The world is not ready for that, I'm afraid. RHEL 8 has Python 3.6 and is not EOL for some time.

But you are right that setup.py should need an update. We've updated the github workflow file, but forgot about the version list there.

kloczek:
Someone is going to push latest `websockify` to RHEL 8 ?
That would be a bit odd ..
AFAIK now RGEL 8 provides only critical fixes.
Does just released version provides such fixes? 🤔 
If yes someone from REL can backport those changes to hey version packaged as rpm package. Isn't it? 🤔

CendioOssman:
Users should be able to manually install websockify, even the latest version, on current distributions. The distributions will unfortunately not keep things updated, so it falls on us.

kloczek:
Thank you 👍

PR #581, opened 2024-07-16
close-all-files uses too much resources

thomasgoirand:
Since Systemd 256~rc3-3, maxfd could be *really* big, and therefore, could take too much resources. This patch attemps to limit this.

### DIFF
--- a/websockify/websockifyserver.py
+++ b/websockify/websockifyserver.py
@@ -521,6 +521,12 @@ class WebSockifyServer():
         # Close open files
         maxfd = resource.getrlimit(resource.RLIMIT_NOFILE)[1]
         if maxfd == resource.RLIM_INFINITY: maxfd = 256
+        # Since Systemd 256~rc3-3, maxfd could be
+        # *really* big, and therefore, the below code
+        # could take too much resources. This somehow
+        # attemps to limit this.
+        if maxfd > 4096:
+            maxfd = 4096
         for fd in reversed(range(maxfd)):
             try:
                 if fd not in keepfd:


CendioOssman:
Thanks for your suggested fix!

I'm a bit cautious about just picking an arbitrary limit, though. Perhaps it's better to move away from this brute force method and instead enumerate open file descriptors by looking in `/proc/self/fd` instead?

thomasgoirand:
Hi.
Indeed, I would have like to do it better as well, but didn't find out how to enumerate the open FDs in a nice way. There's also the os.closeall() function that may be used too, maybe it's possible to just call it with the lowest (3) and highest value possible in the system, and let it deal with it? I honestly don't have enough time to investigate. I'm only a package maintainer in Debian, busy packaging all of OpenStack, Ceph, OpenVSwitch, RabbitMQ and so on, so there's only a limited amount of time I can invest in each bug. The current patch I used just does the job: the package isn't stuck in a loop doing tests. So I'll let you investigate a better solution that I'll adopt as soon as you release it. Thanks for your understanding.

Fabian-Gruenbichler:
> I assume you mean `os.closerange()`? It should have similar issues. Might be fast enough, since it's likely written in C instead. But might not.

this came up again recently (in the context of infra hosts being upgraded to Debian Trixie, which has a higher hard nofile rlimit by default). os.closerange() should be fast, since it uses the `close_range` syscall by default - for the new default features there's about a factor of 100 between:

```
#!/bin/python3

import errno
import os
import resource
import sys

maxfd = resource.getrlimit(resource.RLIMIT_NOFILE)[1]

print(range(maxfd))

for fd in reversed(range(maxfd)):
    try:
        os.close(fd)
    except OSError:
        _, exc, _ = sys.exc_info()
        if exc.errno != errno.EBADF: raise
```

and

```
#!/bin/python3

import os
import resource

maxfd = resource.getrlimit(resource.RLIMIT_NOFILE)[1]

print(range(maxfd))

os.closerange(0,maxfd)
```

it does eat errors though..

CendioOssman:
Did you also check the suggested patch above?

PR #582, opened 2024-07-18
Make dependencies for Token plugins optional

javacruft:
redis and jwcrypto are not required to use websockify so configure these as optional requirements for specific features (redis, jwt) rather than always installing.


CendioOssman:
Thank you for your PR!

Yes, those are optional. But might be overlooked by users, so there is some value to not forcing users to jump through extra hoops.

What issues have you seen by having redis and jwcrypto also getting installed?

PR #601, opened 2025-03-12
Add lint workflow

### LINKED ISSUE #600: Add linter to our workflow.
Right now, we aren't checking for simple syntax errors when we run our tests.

Such errors should *optimally* be found when running the unittests. But if the tests do not reach a specific code path for some reason, a lint check should serve as a good sanity check of the state of the code.

ThirVondukr:
Closes #600 

This PR adds:

### Package Manager
https://docs.astral.sh/uv/
Tool to manage project dependencies, with support for `pyproject.toml` (https://peps.python.org/pep-0621/)

### Mypy
https://github.com/python/mypy
Static type checker

### Ruff
https://docs.astral.sh/ruff/
Linter and formatter

`workflows/lint.yaml` uses python 3.12, I don't think this should be an issue as long as target versions are configured properly for ruff and mypy.


ThirVondukr:
Latest versions of mypy and ruff only support python 3.8 and 3.7 respectively 🤔

d0tiKs:
> Latest versions of mypy and ruff only support python 3.8 and 3.7 respectively 🤔

> **[What versions of Python does Ruff support?](https://docs.astral.sh/ruff/faq/#what-versions-of-python-does-ruff-support)**
> Ruff can **lint code** for any Python version from **3.7 onwards**, including Python 3.13.
> [...]
> Ruff is **installable** under any Python version from **3.7 onwards**.



And for mypy it seems to be : `Requires: Python >=3.9`
https://pypi.org/project/mypy/

ThirVondukr:
> Comparing to novnc compatibility you mean ?
Yes, currently latest release of websockify on pypi states it's compatible with 3.4 and on master branch it's 3.6.
As for mypy - it's possible to use slightly older version, I think that's already handled in `uv.lock`:

```
[package.dev-dependencies]
dev = [
    { name = "mypy", version = "0.971", ..., marker = "python_full_version < '3.7'" },
    { name = "mypy", version = "1.4.1", ..., marker = "python_full_version == '3.7.*'" },
    { name = "mypy", version = "1.14.1", ..., marker = "python_full_version == '3.8.*'" },
    { name = "mypy", version = "1.15.0", ..., marker = "python_full_version >= '3.9'" },
]
```
only quite early version of ruff (`0.0.17`) supports 3.6 though

ThirVondukr:
Also it seems like only ~3% of downloads are for python 3.6 and 3.7 https://pypistats.org/packages/websockify

ThirVondukr:
@ThinLinc-Zeijlon Sorry to ping you, but I'm not quite sure who maintains the repository right now, but I see that the last commit/release was made by you.  
Would you have a bit of time to glance over this PR? Last commit (the one that runs `ruff format`) can probably be rolled back to reduce review time, and one of the maintainers could run it locally.

ThinLinc-Zeijlon:
@ThirVondukr, Thank you for your patience, I'll take a look at your changes :slightly_smiling_face:

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.12
+        job:
+          - mypy .
+          - ruff format --check .
+          - ruff check .
```
ThinLinc-Zeijlon, line 14: We'd like to maintain support for older systems. Could you check how far we could get with something more tried and tested like flake8 and pylint?

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install uv
```
ThinLinc-Zeijlon, line 23: Same as above.

setup.py
ThinLinc-Zeijlon, line 1: Shouldn't setup.py be removed if we are converting to pyproject.toml?

tests/echo.py
```diff
@@ -33,9 +36,11 @@ def new_websocket_client(self):
         while True:
             wlist = []
 
-            if cqueue or c_pend: wlist.append(self.request)
+            if cqueue or c_pend:
+                wlist.append(self.request)
```
ThinLinc-Zeijlon, line 40: This is a nice improvement!

tests/echo.py
```diff
@@ -1,23 +1,26 @@
 #!/usr/bin/env python
 
-'''
```
ThinLinc-Zeijlon, line 3: I think this is a bit too strict. Can the linter be configured to ignore these types of changes?

tests/echo.py
```diff
@@ -50,20 +55,27 @@ def new_websocket_client(self):
                 if closed:
                     break
 
-if __name__ == '__main__':
```
ThinLinc-Zeijlon, line 53: In general, we want the choice of quotation characters to be up the developer.

tests/echo.py
```diff
@@ -50,20 +55,27 @@ def new_websocket_client(self):
...
+        action="store_true",
+        help="verbose messages and per frame traffic",
+    )
+    parser.add_option("--cert", default="self.pem", help="SSL certificate file")
+    parser.add_option(
+        "--key", default=None, help="SSL key file (if separate from cert)"
+    )
+    parser.add_option(
+        "--ssl-only", action="store_true", help="disallow non-encrypted connections"
+    )
```
ThinLinc-Zeijlon, line 73: I think I can guess the strategy it had when formatting here, but I think this made the code less readable.

ThinLinc-Zeijlon (changes requested):

.gitignore
```diff
@@ -10,3 +10,6 @@ target.cfg.d
 build/
 dist/
 *.egg-info
+
+# Intellij IDEs
+.idea
```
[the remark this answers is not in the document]
ThirVondukr: I thought intellij-based ide's are kind of common and adding that to `.gitignore` made sense, so contributors don't have do do that themselves. I can remove that if you want though.

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+readme = "README.md"
+authors = [
+    { name = "Joel Martin", email = "github@martintribe.org" }
+]
+requires-python = ">=3.6"
+dependencies = [
+    "jwcrypto>=1.5.1",
+    "numpy>=1.19.0",
+    "redis>=4.3.0",
+    "requests>=2.27.0",
```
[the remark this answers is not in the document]
ThirVondukr: They're probably the latest compatible releases with 3.7, I don't quite remember if I pulled these manually or uv did it.

uv.lock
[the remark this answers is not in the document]
ThirVondukr: I think lock file is fine to have in a repo. Usually if the project is deployed somewhere (which isn't the case here) you'd want all builds to be reproducible, so lock file helps with that, but it can be used for CI there. (Also you won't really want for dev dependencies to be "automatically" updated either).

.github/workflows/lint.yaml
ThirVondukr: I didn't use these for quite a while, and I didn't use flake8 since migrating to ruff around a year ago. It should include majority of flake8 and some of the pylint checks though.

tests/echo.py
ThirVondukr: Yes, that can be configured to preserve quotes, but the default style is using `"`, I personally prefer it that way.
https://docs.astral.sh/ruff/settings/#format_quote-style

ThirVondukr: Personally I don't think it matters that much, having one quote style is better in my opinion, plus developers still can use whatever quote character they want, it would just get reformatted later, but choice is up to you on that.

tests/echo.py
ThirVondukr: I think generally you'd want that code to be either on one line:
```py
parser.add_option("--verbose", "-v", action="store_true", help="verbose messages and per frame traffic")
```
or like this:
```py
parser.add_option(
    "--verbose",
    "-v",
    action="store_true",
    help="verbose messages and per frame traffic",
)
```
To force the second style you'd need to add `,` to the last argument, but I think formatter could be disabled for certain code blocks: https://docs.astral.sh/ruff/formatter/#format-suppression

CendioZeijlon:
@ossman and @samhed, what are your thoughts on adding `ruff` as a linter for us?

Issue #603, opened 2025-05-17
Dockerfile can not be built soley with the repo


Checker8763:
The Docker image can not be build with the untouched/cloned repository.

This is not so nice because the for automatic build systems like komodo the repo can not easily be set a the base for the image, because running the setup.py is needed.

I will provide a PR with changes to the Dockerfile namely a build stage that runs the setup.py.

PR #604, opened 2025-05-17
Enable docker build by only cloning the repository

Checker8763:
Fixes "Dockerfile can not be built soley with the repo" #603

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,17 @@
-FROM python
+FROM python:3-slim AS builder
 
-COPY websockify-*.tar.gz /
+WORKDIR /build/websockify
 
-RUN python3 -m pip install websockify-*.tar.gz
-RUN rm -rf /websockify-* /root/.cache
+COPY ./* /build/websockify
+
+RUN pip3 install setuptools && \
+    python3 setup.py sdist --dist-dir docker/
+
+# Actual final image
+FROM python:3-slim
+COPY --from=builder /build/websockify/websockify-*.tar.gz /
+RUN python3 -m pip install websockify-*.tar.gz && \
+  rm -rf /websockify-* /root/.cache
 
 VOLUME /data
 
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(name=name,
       author="Joel Martin",
       author_email="github@martintribe.org",
 
-      packages=['websockify'],
+      packages=find_packages(),
       include_package_data=True,
       install_requires=[
         'numpy', 'requests',


setup.py
```diff
@@ -28,7 +28,7 @@
       author="Joel Martin",
       author_email="github@martintribe.org",
 
-      packages=['websockify'],
+      packages=find_packages(),
```
CendioOssman, line 31: Not sure this is a plus, given we have such a simple project. Explicitly listing them gives you some protection against including any random junk that people might have in their working copy.

docker/Dockerfile
```diff
@@ -1,9 +1,17 @@
...
+COPY ./* /build/websockify
+
+RUN pip3 install setuptools && \
+    python3 setup.py sdist --dist-dir docker/
+
+# Actual final image
+FROM python:3-slim
+COPY --from=builder /build/websockify/websockify-*.tar.gz /
+RUN python3 -m pip install websockify-*.tar.gz && \
+  rm -rf /websockify-* /root/.cache
```
CendioOssman, line 14: What is it that gets cleaned up in this new final line?

docker/Dockerfile
```diff
@@ -1,9 +1,17 @@
...
 
-RUN python3 -m pip install websockify-*.tar.gz
-RUN rm -rf /websockify-* /root/.cache
+COPY ./* /build/websockify
+
+RUN pip3 install setuptools && \
+    python3 setup.py sdist --dist-dir docker/
+
+# Actual final image
+FROM python:3-slim
```
CendioOssman, line 11: What are we losing in this switch? We have a bunch of optional dependencies, like numpy, that users probably want.

CendioOssman (changes requested):
Thanks! This looks promising.

setup.py
CendioOssman: Please do and share the error here.

docker/Dockerfile
Checker8763: I have found the following section in the python docs:

~ https://docs.python.org/3.11/distutils/setupscript.html#listing-whole-packages

```
The keys to this dictionary are package names, 
and an empty package name stands for the root package. 
The values are directory names relative to your distribution root. 
In this case, when you say packages = ['foo'], 
you are promising that the file lib/foo/__init__.py exists.
```

Checker8763: While reading in https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

Another workaround I have found was to include `package_dir = {"websockify": ""},` or `package_dir = {"websockify": "."},` or `package_dir = {"websockify": "/"},`

This seems to be the least required amount of changes, to make the dockerfile work while leaving the statements in the dockerfile unchanged.

setup.py
Checker8763: https://github.com/novnc/websockify/pull/604#discussion_r2134664805

I posted into the wrong comment...

docker/Dockerfile
CendioOssman: I'm afraid I cannot reproduce the issue here. And it really should work. The documentation you refer to states:

> Thus, when you say packages = ['foo'] in your setup script, you are promising that the Distutils will find a file foo/\_\_init\_\_.py (which might be spelled differently on your system, but you get the idea) relative to the directory where your setup script lives.

And since we say `packages=['websockify']`, we claim that there is a `websockify` directory next to `setup.py`. Which there is.

It seems like there is something more in your environment that is fouling thing.

PR #606, opened 2025-06-03
Fix outdated tox.ini

kajinamit:
Update the default environment and test command so that we can run unit tests by the tox command properly.


CendioZeijlon:
Looks good, thanks for contributing!

PR #607, opened 2025-06-03
Require Python >= 3.6

kajinamit:
Python 3.4 and 3.5 haven't been tested for long time. Enforce the minimum version so that users may not attempt to use untested ancient python versions.

Note that Python 3.6 ~ 3.8 already reached their EOL and we may even remove support for these in the near future.


CendioZeijlon:
I agree with you on this. I think one good example here is that one could easily break backward compatibility just by using `f{}` format strings, which was added in Python 3.6.

CendioOssman:
Yes, I'd say we're formally on 3.6 since 3f205f7cc6131368631c8b0cb17c3aa5689507e9. The changes in this PR should probably have been included at the same time. Thanks for noticing and fixing it!

As for dropping 3.6, that's only EOL from upstream Python. It is still used in RHEL 8, which isn't EOL until 2029 at the earliest.

PR #608, opened 2025-06-04
Remove run script alias

kajinamit:
It was added 13 years ago for compatibility when the run script was converted to a shell script by [1]. The long time has passed since then and we can assume that users may no longer use it. Removing it allows us to resolve the confusing name (.py which triggers shell).

[1] 4f6a82159ce42299c8666e04f4a4c3d0fe829803


CendioOssman:
We might as well remove it in that case. It was added as a backward compatibility in c0d23e27e4b94ee549d3dd5918fdc89bd91171ca.

PR #609, opened 2025-06-04
Migrate to modern logger interface

emmanuel-ferdman:
## Description
The `logger.warn()` method is deprecated since Python2.7 and replaced with `logger.warning()`. It leads to those warnings:
```python
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```
This small PR resolves those warnings.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -439,8 +439,8 @@ def select_ssl_version(version):
         keys.sort()
         fallback = keys[-1]
         logger = logging.getLogger(WebSocketProxy.log_prefix)
-        logger.warn("TLS version %s unsupported. Falling back to %s",
-                    version, fallback)
+        logger.warning("TLS version %s unsupported. Falling back to %s",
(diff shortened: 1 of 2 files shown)


kajinamit (approved):
LGTM. Grep shows no other usage of warn.

emmanuel-ferdman:
@CendioZeijlon `logging.WARN` is not deprecated like `logging.warn`, just simple alias. Should I still replace `logging.WARN` with `logging.WARNING`?

CendioZeijlon:
> logging.WARN is not deprecated like logging.warn, just simple alias. Should I still replace logging.WARN with logging.WARNING?

My guess is that WARN was kept as an alias to not break backward compatibility when renaming the variable to WARNING. Python's docs seem to only mention `logging.WARNING`, so I figured this could be a good opportunity to make the change to be consistent with the docs.

emmanuel-ferdman:
@CendioZeijlon sure, done 😄

PR #610, opened 2025-06-08
Fix flake8 errors

kajinamit:
Fix flake8 errors to improve readability of the code and also avoid potential bugs.

### DIFF
--- a/tests/load.py
+++ b/tests/load.py
@@ -43,8 +43,6 @@ class WebSocketLoad(WebSockifyRequestHandler):
 
     def responder(self, client):
         c_pend = 0
-        cqueue = []
-        cpartial = ""
         socks = [client]
         last_send = time.time() * 1000
 
--- a/tests/load.py
+++ b/tests/load.py
@@ -76,11 +76,7 @@ class WebSocketLoad(WebSockifyRequestHandler):
     def generate(self):
         length = random.randint(10, self.max_packet_size)
         numlist = self.rand_array[self.max_packet_size - length:]
-        # Error in length
-        #numlist.append(5)
         chksum = sum(numlist)
-        # Error in checksum
-        #numlist[0] = 5
         nums = "".join([str(n) for n in numlist])
         data = "^%d:%d:%d:%s$" % (self.send_cnt, length, chksum, nums)
         self.send_cnt += 1
--- a/tests/load.py
+++ b/tests/load.py
@@ -133,13 +133,13 @@ class WebSocketLoad(WebSockifyRequestHandler):
 if __name__ == '__main__':
     parser = optparse.OptionParser(usage="%prog [options] listen_port")
     parser.add_option("--verbose", "-v", action="store_true",
-            help="verbose messages and per frame traffic")
+                      help="verbose messages and per frame traffic")
     parser.add_option("--cert", default="self.pem",
-            help="SSL certificate file")
+                      help="SSL certificate file")
     parser.add_option("--key", default=None,
-            help="SSL key file (if separate from cert)")
+                      help="SSL key file (if separate from cert)")
     parser.add_option("--ssl-only", action="store_true",
-            help="disallow non-encrypted connections")
+                      help="disallow non-encrypted connections")
     (opts, args) = parser.parse_args()
 
     try:
--- a/tests/load.py
+++ b/tests/load.py
@@ -7,7 +7,7 @@ given a sequence number. Any errors are reported and counted.
 '''
 
 import sys, os, select, random, time, optparse, logging
-sys.path.insert(0,os.path.join(os.path.dirname(__file__), ".."))
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 from websockify.websockifyserver import WebSockifyServer, WebSockifyRequestHandler
 
 
@@ -74,13 +74,13 @@ class WebSocketLoad(WebSockifyRequestHandler):
 
     def generate(self):
         length = random.randint(10, self.max_packet_size)
-        numlist = self.rand_array[self.max_packet_size-length:]
+        numlist = self.rand_array[self.max_packet_size - length:]
         # Error in length
         #numlist.append(5)
         chksum = sum(numlist)
         # Error in checksum
         #numlist[0] = 5
-        nums = "".join( [str(n) for n in numlist] )
+        nums = "".join([str(n) for n in numlist])
         data = "^%d:%d:%d:%s$" % (self.send_cnt, length, chksum, nums)
         self.send_cnt += 1
 
@@ -102,7 +102,7 @@ class WebSocketLoad(WebSockifyRequestHandler):
 
             try:
                 cnt, length, chksum, nums = data[1:-1].split(':')
-                cnt    = int(cnt)
+                cnt = int(cnt)
                 length = int(length)
                 chksum = int(chksum)
             except ValueError:
@@ -151,7 +151,7 @@ if __name__ == '__main__':
         if len(args) != 1: raise ValueError
         opts.listen_port = int(args[0])
 
-        if len(args) not in [1,2]: raise ValueError
+        if len(args) not in [1, 2]: raise ValueError
         opts.listen_port = int(args[0])
         if len(args) == 2:
             opts.delay = int(args[1])
--- a/tests/load.py
+++ b/tests/load.py
@@ -10,6 +10,7 @@ import sys, os, select, random, time, optparse, logging
 sys.path.insert(0,os.path.join(os.path.dirname(__file__), ".."))
 from websockify.websockifyserver import WebSockifyServer, WebSockifyRequestHandler
 
+
 class WebSocketLoadServer(WebSockifyServer):
 
     recv_cnt = 0
@@ -84,7 +85,6 @@ class WebSocketLoad(WebSockifyRequestHandler):
         self.send_cnt += 1
 
         return data
-
 
     def check(self, frames):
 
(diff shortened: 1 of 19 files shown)


kajinamit:
Note: Most of the fixes (especially fixes about blank line and white space errors) are generated by autopep8

CendioOssman (changes requested):
Nice cleanups!

Have you explored getting this in to GitHub actions as well? So we don't regress.

websockify/websocketproxy.py
```diff
@@ -293,20 +293,20 @@ class WebSocketProxy(websockifyserver.WebSockifyServer):
...
-        self.unix_target    = kwargs.pop('unix_target', None)
-        self.ssl_target     = kwargs.pop('ssl_target', None)
-        self.heartbeat      = kwargs.pop('heartbeat', None)
+        self.target_host = kwargs.pop('target_host', None)
+        self.target_port = kwargs.pop('target_port', None)
+        self.wrap_cmd = kwargs.pop('wrap_cmd', None)
+        self.wrap_mode = kwargs.pop('wrap_mode', None)
+        self.unix_target = kwargs.pop('unix_target', None)
+        self.ssl_target = kwargs.pop('ssl_target', None)
+        self.heartbeat = kwargs.pop('heartbeat', None)
```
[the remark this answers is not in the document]
kajinamit: The bare approach might be adding `noqa:` comment to each line. We can disable the check for the whole file but IMO that is not ideal.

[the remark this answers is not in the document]
kajinamit: OK I think I could address this.

CendioOssman (changes requested):

CendioOssman:
> I've added the workflow to run pep8 check. There are still number of checks disabled but at least we can detect any new errors. Does it make sense to you ?

Sounds great! It allows us to lock down what we already agree on, and explore other things as we want/need.

tests/test_websockifyserver.py
```diff
@@ -344,7 +344,6 @@ def test_fallback_sigchld_handler(self):
 
     def test_start_server_error(self):
         server = self._get_server(daemon=False, ssl_only=1, idle_timeout=1)
-        sock = server.socket('localhost')
```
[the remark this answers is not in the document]
kajinamit: Updated accordingly

websockify/token_plugins.py
```diff
@@ -371,5 +371,5 @@ def lookup(self, token):
 
             return ['unix_socket', uds_path]
         except Exception as e:
-                logger.error("Error finding unix domain socket: %s" % str(e))
-                return None
+            logger.error("Error finding unix domain socket: %s" % str(e))
+            return None
```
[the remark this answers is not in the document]
kajinamit: Makes sense. Updated accordingly

tests/load.py
```diff
@@ -43,8 +43,8 @@ def new_websocket_client(self):
 
     def responder(self, client):
         c_pend = 0
-        cqueue = []
-        cpartial = ""
+        cqueue = []  # noqa: F841
+        cpartial = ""  # noqa: F841
```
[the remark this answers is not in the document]
kajinamit: Done

tests/test_token_plugins.py
```diff
@@ -207,7 +207,7 @@ def test_asymmetric_jwe_token_plugin(self):
 class TokenRedisTestCase(unittest.TestCase):
     def setUp(self):
         try:
-            import redis
+            import redis  # noqa: F401
```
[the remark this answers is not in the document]
kajinamit: I've moved this import to the top level.

tox.ini
```diff
@@ -15,3 +15,7 @@ deps = -r{toxinidir}/test-requirements.txt
 [testenv:pep8]
 commands = flake8
 deps = flake8
+
+[flake8]
+ignore = E126,E127,E128,E129,E401,E402,W504
+max-line-length = 160
```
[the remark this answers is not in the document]
kajinamit: OK I managed to get rid of all ignores.

websockify/websocketproxy.py
```diff
@@ -298,15 +298,15 @@ def __init__(self, RequestHandlerClass=ProxyRequestHandler, *args, **kwargs):
...
-        self.ssl_target = kwargs.pop('ssl_target', None)
-        self.heartbeat = kwargs.pop('heartbeat', None)
+        self.ssl_target  = kwargs.pop('ssl_target', None)  # noqa: E221
+        self.heartbeat   = kwargs.pop('heartbeat', None)  # noqa: E221
 
         self.token_plugin = kwargs.pop('token_plugin', None)
-        self.host_token = kwargs.pop('host_token', None)
-        self.auth_plugin = kwargs.pop('auth_plugin', None)
+        self.host_token   = kwargs.pop('host_token', None)  # noqa: E221
+        self.auth_plugin  = kwargs.pop('auth_plugin', None)  # noqa: E221
```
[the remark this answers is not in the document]
kajinamit: It has been a while since we discussed this. I wonder if we can agree with following the pep8 guideline now (I mean, remove the adjustment commit for now) and "fix" these later if we want.
I know this is not urgent but I hope we can merge this soon, because this likely conflict with any new change, due to multiple lines being updated.

websockify/websocketserver.py
```diff
@@ -64,7 +64,7 @@ def _websocket_do_GET(self):
         self.do_GET = self._real_do_GET
 
         if (self.headers.get('upgrade') and
-            self.headers.get('upgrade').lower() == 'websocket'):
+                self.headers.get('upgrade').lower() == 'websocket'):
```
CendioOssman: This change doesn't seem to be required by PEP8?

websockify/sysloghandler.py
```diff
@@ -110,5 +110,5 @@ def emit(self, record):
 
         except (KeyboardInterrupt, SystemExit):
             raise
-        except:
+        except Exception:
```
CendioOssman: Looks like we can remove the special handling of `KeyboardInterrupt` and `SystemExit` with this change?

CendioOssman (changes requested):
Thanks for all the adjustments! I think this is pretty much ready to go.

There was one indentation change that looked strange, though. Could you have a look at that, after which we can merge this.

websockify/websocketserver.py
kajinamit: No. This is required to address E125 Error

```
./websockify/websocketserver.py:67:13: E129 visually indented line with same indent as next logical line
```

websockify/sysloghandler.py
kajinamit: Oh, yes. Because these two don't inherit Exception, these should not be caught here and are raised automatically.

kajinamit:
Thanks for the additional feedback !
The indentation change is what the E125 rule requires. Please see my inline comment.

websockify/websocketserver.py
CendioOssman: This is controversial and not in alignment with PEP8:

https://github.com/PyCQA/pycodestyle/issues/126

So it apparently got split out to E129, rather than being grouped in with the okay things in E125.

But it's still enabled by default, which is also (understandably) controversial:

https://github.com/PyCQA/pycodestyle/issues/386

It would be nice if we could avoid having exceptions. But I don't want people fighting the linter either.

Since this is permitted in PEP8, I think the lesser evil is disabling E129 globally.

kajinamit: OK. I disabled the check and reverted the change.

kajinamit:
~Hmm it's wired that tox.ini is ignored in CI. I can't reproduce the issue in my local...~

Ignore this. I noticed I didn't commit the additional change in tox.ini properly.

tox.ini
```diff
@@ -18,3 +18,5 @@ deps = flake8
 
 [flake8]
 max-line-length = 160
+# E129 visually indented line with same indent as next logical line
+ignore = E121,E123,E126,E226,E24,E704,W503,W504,E129
```
CendioOssman: Something went wrong here, this is not just E129?

CendioOssman (changes requested):

websockify/websocketserver.py
```diff
@@ -64,7 +64,7 @@ def _websocket_do_GET(self):
         self.do_GET = self._real_do_GET
 
         if (self.headers.get('upgrade') and
-                self.headers.get('upgrade').lower() == 'websocket'):
+            self.headers.get('upgrade').lower() == 'websocket'):
```
CendioOssman: And please revert this change in the earlier commit, rather than changing it back and forth.

CendioOssman (changes requested):

tox.ini
kajinamit: Adding ignore = E129 clears the checks disabled by default in flake8. I added E129 in addition to what are disabled by default.

kajinamit: Updated the line to explicitly disable W504 (and W503 tightly coupled with W504), because the other checks are not failing. Maybe this is cleaner though it enables a few checks (eg. E121) which are disabled by default.

websockify/websocketserver.py
kajinamit: Fixed accordingly.

tox.ini
CendioOssman: > Adding ignore = E129 clears the checks disabled by default in flake8. I added E129 in addition to what are disabled by default.

I see. What a confusing behaviour.

> Updated the line to explicitly disable W504 (and W503 tightly coupled with W504), because the other checks are not failing. Maybe this is cleaner though it enables a few checks (eg. E121) which are disabled by default.

Let's try it. We can disable them again later if we find them annoying. :)

CendioOssman (approved):
Looks great!

Thanks for doing all the hard work here!

PR #611, opened 2025-06-12
Return 405 Method Not Allowed in request handler

tobias-urdin:
Instead of returning HTTP 501 lets implement the
other HTTP methods and returns a HTTP 405 Method
Not Allowed.

### DIFF
--- a/tests/test_websockifyserver.py
+++ b/tests/test_websockifyserver.py
@@ -89,6 +89,60 @@ class WebSockifyRequestHandlerTestCase(unittest.TestCase):
         send_error.assert_called_with(405)
 
     @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_error')
+    def test_head_with_only_upgrade_returns_error(self, send_error):
+        server = self._get_server(web=None)
+        handler = websockifyserver.WebSockifyRequestHandler(
+            FakeSocket(b'HEAD /tmp.txt HTTP/1.1'), '127.0.0.1', server)
+
+        handler.do_HEAD()
+        send_error.assert_called_with(405)
+
+    @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_error')
+    def test_post_returns_error(self, send_error):
+        server = self._get_server(web=None)
+        handler = websockifyserver.WebSockifyRequestHandler(
+            FakeSocket(b'POST / HTTP/1.1'), '127.0.0.1', server)
+
+        handler.do_POST()
+        send_error.assert_called_with(405)
+
+    @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_error')
+    def test_put_returns_error(self, send_error):
+        server = self._get_server(web=None)
+        handler = websockifyserver.WebSockifyRequestHandler(
+            FakeSocket(b'PUT / HTTP/1.1'), '127.0.0.1', server)
+
+        handler.do_PUT()
+        send_error.assert_called_with(405)
+
+    @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_error')
+    def test_patch_returns_error(self, send_error):
+        server = self._get_server(web=None)
+        handler = websockifyserver.WebSockifyRequestHandler(
+            FakeSocket(b'PATCH / HTTP/1.1'), '127.0.0.1', server)
+
+        handler.do_PATCH()
+        send_error.assert_called_with(405)
+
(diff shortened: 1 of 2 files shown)


CendioZeijlon (changes requested):

tobias-urdin:
Updated again per your comments :)

tests/test_websockifyserver.py
```diff
@@ -79,14 +79,86 @@ def _get_server(self, handler_class=websockifyserver.WebSockifyRequestHandler,
...
+    def test_normal_get_with_only_upgrade_returns_error(
+        self, send_error, send_hdr
+    ):
         server = self._get_server(web=None)
         handler = websockifyserver.WebSockifyRequestHandler(
             FakeSocket(b'GET /tmp.txt HTTP/1.1'), '127.0.0.1', server)
 
         handler.do_GET()
         send_error.assert_called_with(405)
+        send_hdr.assert_not_called()
```
CendioZeijlon, line 93: It's better to not make this assumption, e.g. if a future change adds a `send_hdr()`-call that's unrelated to the Accept header. Let's just remove this line.

tests/test_websockifyserver.py
```diff
@@ -79,14 +79,86 @@ def _get_server(self, handler_class=websockifyserver.WebSockifyRequestHandler,
...
+    def test_head_with_only_upgrade_returns_error(
+        self, send_error, send_hdr
+    ):
+        server = self._get_server(web=None)
+        handler = websockifyserver.WebSockifyRequestHandler(
+            FakeSocket(b'HEAD /tmp.txt HTTP/1.1'), '127.0.0.1', server)
+
+        handler.do_HEAD()
+        send_error.assert_called_with(405)
+        send_hdr.assert_not_called()
```
CendioZeijlon, line 106: Remove this line as well!

tests/test_websockifyserver.py
```diff
@@ -79,14 +79,86 @@ def _get_server(self, handler_class=websockifyserver.WebSockifyRequestHandler,
             record=self.tmpdir, daemon=False, ssl_only=0, idle_timeout=1,
             **kwargs)
 
+    @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_header')
     @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_error')
-    def test_normal_get_with_only_upgrade_returns_error(self, send_error):
+    def test_normal_get_with_only_upgrade_returns_error(
+        self, send_error, send_hdr
```
CendioZeijlon, line 85: Remove send_hdr when it's not used anymore.

tests/test_websockifyserver.py
```diff
@@ -79,14 +79,86 @@ def _get_server(self, handler_class=websockifyserver.WebSockifyRequestHandler,
...
             FakeSocket(b'GET /tmp.txt HTTP/1.1'), '127.0.0.1', server)
 
         handler.do_GET()
         send_error.assert_called_with(405)
+        send_hdr.assert_not_called()
+
+    @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_header')
+    @patch('websockify.websockifyserver.WebSockifyRequestHandler.send_error')
+    def test_head_with_only_upgrade_returns_error(
+        self, send_error, send_hdr
```
CendioZeijlon, line 98: Remove send_hdr when it's not used anymore. (See other comments about `send_hdr.assert_not_called()`)

CendioZeijlon (changes requested):
This is starting to look good, just a few minor changes left :)

tests/test_websockifyserver.py
tobias-urdin: Fixed

tests/test_websockifyserver.py
tobias-urdin: Fixed

tobias-urdin:
> I tested running `requests.post('<hostname>', data={})` against a running server. I am getting
> 
> ```
> requests.exceptions.ConnectionError: ('Connection aborted.', BadStatusLine('Accept: HEAD, GET\r\n'))
> ```
> 
> Looks like `send_error()` builds a complete response, and isn't meant to be used with `send_header()` or any of the methods that can be used to build a custom response...
> 
> So the simplest way forward is to revert the latest changes and do what you did from the beginning, to just call `self.send_error(405)`. Sorry about that.

Removed ``send_header()``, let me know if there is any other way to send a header otherwise we'll leave it for the future.

PR #616, opened 2025-07-03
Support for URI prefix

kwkoo:
When specified, the web server will serve files off the specified URI. For example, if websockify is launched with "--prefix vnc", then users will need to access from the URL http://server/vnc/.

This is useful in deployments where a reverse proxy has multiple upstreams, with websockify being one of the upstream servers.

### DIFF
--- a/websockify/websocketproxy.py
+++ b/websockify/websocketproxy.py
@@ -482,6 +482,8 @@ def websockify_init():
                  "active connections")
     parser.add_option("--cert", default="self.pem",
             help="SSL certificate file")
+    parser.add_option("--prefix", default=None,
+            help="URI prefix")
     parser.add_option("--key", default=None,
             help="SSL key file (if separate from cert)")
     parser.add_option("--key-password", default=None,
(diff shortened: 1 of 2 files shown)


CendioOssman:
This seems a bit convoluted. If you have a reverse proxy, can't you just have it adjust the path? I'd say that's the norm when having multiple web applications behind a single host.

kwkoo:
Thanks for the comment.

It's true that most reverse proxies let you rewrite the URI. However, things get a lot more complicated when `websockify` is behind a chain of reverse proxies.

In my case, the deployment looks like this:
`web server` -> `oauth2-proxy` -> `websockify`

If I configure the web server to rewrite `/vnc` to `/`, then oauth2-proxy will attempt to redirect the user to `/` after the user has successfully authenticated.

If I set websockify to serve file with the `/vnc` prefix, then the web server doesn't have to rewrite the URI, and `oauth2-proxy` will redirect the user to `/vnc` after the user has successfully authenticated and everything works as intended.

CendioOssman:
That sounds like an issue with the oauth proxy in that case.

I'm afraid I'm going to say no to this at this time. Having fixed URLs is something I'd consider normal behaviour from a web application.

As a workaround, you could put a symlink in the noVNC directory, e.g. `ln -s . vnc`, to get the same effect.

PR #622, opened 2025-09-28
Fix FTBFS with GCC 15

thomasgoirand:
Without this patch, websockify fails to build from source.


CendioZeijlon:
Looks great, thanks for contributing!

PR #624, opened 2025-11-03
Added ipv6 support for command wrapping

CraftingDragon007:
I updated the rebind.c file to also support ipv6. This allows wrapping of programs listening on ipv6 (for example: [::]:5900 or [::1]:5900)

### DIFF
--- a/rebind.c
+++ b/rebind.c
@@ -26,7 +26,9 @@
 #include <dlfcn.h>
 
 #include <string.h>
+#include <arpa/inet.h>
 #include <netinet/in.h>
+#include <sys/socket.h>
 
 
 #if defined(DO_DEBUG)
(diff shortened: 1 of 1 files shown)


CendioZeijlon:
Thanks for contributing! I'll take a look.

rebind.c
```diff
@@ -37,25 +39,36 @@
 #define DEBUG(...)
 #endif
 
-
 int bind(int sockfd, const struct sockaddr *addr, socklen_t addrlen)
 {
-    static void * (*func)(int, const struct sockaddr *, socklen_t);
+    static int (*func)(int, const struct sockaddr *, socklen_t);
```
CendioZeijlon, line 44: Later on, in the old code, the return value is cast to a `long`. Are you sure that `int` is sufficient here?

rebind.c
```diff
@@ -37,25 +39,36 @@
...
-          sockfd, addrlen, askaddr, askport);
+    if (family == AF_INET) {
+        const struct sockaddr_in *addr_in = (const struct sockaddr_in *)addr;
+        askport = ntohs(addr_in->sin_port);
+        inet_ntop(AF_INET, &addr_in->sin_addr, addr_str, sizeof(addr_str));
+    } else if (family == AF_INET6) {
+        const struct sockaddr_in6 *addr_in6 = (const struct sockaddr_in6 *)addr;
+        askport = ntohs(addr_in6->sin6_port);
+        inet_ntop(AF_INET6, &addr_in6->sin6_addr, addr_str, sizeof(addr_str));
+    }
```
CendioZeijlon, line 65: Are there other cases we are missing now -> should there be an `else` here?

rebind.c
```diff
@@ -72,22 +85,38 @@ int bind(int sockfd, const struct sockaddr *addr, socklen_t addrlen)
...
+        struct sockaddr_in6 *addr_in6_tmp = (struct sockaddr_in6 *)&addr_tmp;
+        static const struct in6_addr v4_loopback_mapped = { // ::ffff:127.0.0.1
+            .s6_addr = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff, 127, 0, 0, 1}
+        };
+        int v6_only = 0;
+        (void)setsockopt(sockfd, IPPROTO_IPV6, IPV6_V6ONLY, &v6_only, sizeof(v6_only));
+        addr_in6_tmp->sin6_addr = v4_loopback_mapped;
+        addr_in6_tmp->sin6_port = htons(newport);
+        addr_in6_tmp->sin6_scope_id = 0;
+    }
```
CendioZeijlon, line 118: Same question here as above.

CendioZeijlon (commented):
Looks good overall, but it looks like there is both some refactoring and addition of new features.

Could you rework this a bit into separate commits to make your changes easier to follow?

rebind.c
CraftingDragon007: The bind function is specified to return int, so it should be sufficient. [link to documentation](https://www.man7.org/linux/man-pages/man2/bind.2.html#top_of_page)

rebind.c
CraftingDragon007: Yeah I must have overseen that, as there are only ipv4 and ipv6 I could just do else (without if). I will commit it rn

CraftingDragon007:
> Looks good overall, but it looks like there is both some refactoring and addition of new features.

 Yes, I also changed some other things here and there because it was needed for ipv6 (for example storing the ipv6 address)

> Could you rework this a bit into separate commits to make your changes easier to follow?

Does it really make sense to rework changes of +47 -18 lines of code in one single file into multiple commits?

PR #627, opened 2025-12-31
Use new websockify image

bswinnerton:


bswinnerton:
Sorry for the noise! Was targeting the wrong base repo.

PR #634, opened 2026-05-07
Replace sha1 with sha256 to resolve #633

### LINKED ISSUE #633: Replace sha1 with sha256 to satisfy security scanners (CWE-327, CWE-328) and reduce efforts for security analysis
Replace sha1 with sha256 to satisfy security scanners (CWE-327, CWE-328) and reduce efforts for security analysis.

https://github.com/novnc/websockify/blob/master/websockify/websocket.py#L215
```
            expected = sha1((self._key + self.GUID).encode("ascii")).digest()
```

https://github.com/novnc/websockify/blob/master/websockify/websocket.py#L280
```
            # Generate the hash value for the accept header
            accept = sha1((key + self.GUID).encode("ascii")).digest()
```

oneumyvakin:
Replace sha1 with sha256 to satisfy security scanners (CWE-327, CWE-328) and reduce efforts for security analysis

### DIFF
--- a/websockify/websocket.py
+++ b/websockify/websocket.py
@@ -21,7 +21,7 @@ import socket
 import ssl
 import struct
 from base64 import b64encode
-from hashlib import sha1
+from hashlib import sha256
 from urllib.parse import urlparse
 
 try:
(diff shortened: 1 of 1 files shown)


CendioOssman:
SHA-1 is required by the protocol and cannot be changed.  It does not have any security implications:

https://datatracker.ietf.org/doc/html/rfc6455#section-10.8

PR #636, opened 2026-05-17
Bump checkout action

kajinamit:
... to the latest one available.


samhed:
I think v6 is the latest version available

kajinamit:
> I think v6 is the latest version available

Oh yes. I was confused by a different action. Fixed it.

